PR #14, opened 2019-04-04
basic lru with mutex

kk3399:
plan is to add batching on top
details here https://github.com/dgraph-io/ristretto/issues/6

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/14)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2019 Dgraph Labs, Inc. and Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ristretto
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/golang/groupcache/lru"
+)
+
+//Cache interface
+type Cache interface {
+	Get(key []byte) ([]byte, error)
+	Set(key []byte, value []byte) error
+}
+
+//BasicCache implementation
+type BasicCache struct {
+	c    *lru.Cache
+	lock sync.Mutex
+}
+
+//Get a value from cache
+func (r *BasicCache) Get(key []byte) ([]byte, error) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	v, ok := r.c.Get(string(key))
+	if ok {
+		return v.([]byte), nil
+	}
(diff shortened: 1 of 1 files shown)


cache_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"bytes"
+	"testing"
+)
+
+
+func TestGet(t *testing.T){	
+	k := []byte("1")
+	v := []byte("one")
+	cache := New(0)
+	cache.Set(k,v)
```
golangcibot: Error return value of `cache.Set` is not checked (from `errcheck`)

cache.go
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+package ristretto
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/dgraph-io/ristretto/lru"
```
mangalaman93: We could just import it instead of copying it. I don't see any benefits of copying the code right now

cache.go
```diff
@@ -0,0 +1,60 @@
...
+//Cache interface
+type Cache interface {
+	Get(key []byte) ([]byte, error)
+	Set(key []byte, value []byte) error
+}
+
+//BasicCache implementation
+type BasicCache struct {
+	c   *lru.Cache
+	mux sync.Mutex
```
mangalaman93: should just call it `lock`. Mux sometimes refer to http router.

cache_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+ package ristretto
+
+ import (
+	"bytes"
+	"testing"
+)
+
+
+func TestGet(t *testing.T){
```
mangalaman93: LRU code is already tested.
This is really not required.

cache.go
```diff
@@ -0,0 +1,60 @@
...
+//Set a value in cache for a key
+func (r *BasicCache) Set(key, value []byte) error {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	r.c.Add(string(key), value)
+	return nil
+}
+
+//New a BasicCache
+func New(size int) *BasicCache {
```
sourabh1024: Should we change size into options ?
As the functionality of cache grows the parameter size will become cumbersome. Instead we should allow users to set options which can have a default value and only required fields should be outside of options.

sourabh1024: Can be consider for future tasks to limit the scope.

kk3399: added size as option

cache.go
```diff
@@ -55,6 +55,17 @@ func (r *BasicCache) Set(key, value []byte) error {
...
+func New(opts ...func(*BasicCache)) *BasicCache {
+	r := &BasicCache{}
+	for _, opt := range opts {
+		opt(r)
+	}
+	return r
+}
+
+//Size sets size of BasicCache
+func Size(size int) func(*BasicCache) {
```
sourabh1024: For optional fields we should have a default value as well. 
If someone doesn't sets the size will it cause any issues in that case if yes we should have a default value.

mangalaman93:
@kk3399 the PR looks good. We should squash the commits into single commit, and then we will merge it.

manishrjain:
@mangalaman93 -- GitHub allows squash and merge via the interface.

PR #16, opened 2019-04-16
Add utility SLRU implementation

ckarenz:
This adds a basic SLRU implementation meant to roughly match the example set by [github.com/golang/groupcache/lru](https://github.com/golang/groupcache/tree/master/lru).

Design choices:
- I dropped the lazy initialization pattern to skip unnecessary checks each function call.
- I added an Oldest accessor which will be needed for TinyLFU.
- Though I don't strictly adhere to the Cache interface, I did stick with the byte slice keys and values.
- There are likely a few unnecessary string conversions. I expect the map to be replaced, so they should be temporary.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/16)

### DIFF
--- a/slru/slru.go
+++ b/slru/slru.go
@@ -58,7 +58,7 @@ func (c *Cache) Get(key []byte) (value []byte, ok bool) {
 		return
 	}
 
-	// Just promote the entry there's room in the next segment.
+	// Just promote the entry if there's room in the next segment.
 	if c.protected.Len() < c.maxProtected {
 		c.probation.Remove(e)
 		ent.segment = protected
--- a/slru/slru.go
+++ b/slru/slru.go
@@ -1,0 +1,140 @@
+package slru
+
+import (
+	"container/list"
+)
+
+type segment int
+
+const (
+	probation segment = iota
+	protected
+)
+
+// Cache is a segmented LRU cache. It is not safe for concurrent access.
+type Cache struct {
+	items        map[string]*list.Element
+	probation    *list.List
+	protected    *list.List
+	maxProbation int
+	maxProtected int
+}
+
+type entry struct {
+	key     []byte
+	value   []byte
+	segment segment
+}
+
+// New creates a new SLRU cache.
+//
+// Segment capacities must be positive.
+func New(maxProbation int, maxProtected int) *Cache {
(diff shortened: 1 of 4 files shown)


slru/slru.go
```diff
@@ -0,0 +1,140 @@
...
+	probation    *list.List
+	protected    *list.List
+	maxProbation int
+	maxProtected int
+}
+
+type entry struct {
+	key     []byte
+	value   []byte
+	segment segment
```
ckarenz: This could just as easily be a pointer to its containing list, which would be just as expressive, though I worry about giving the GC more work to do in breaking cycles.

slru/slru.go
```diff
@@ -0,0 +1,140 @@
...
+	c.items[keyStr] = prot
+	c.items[string(victim.key)] = e
+	return
+}
+
+// Add adds a value to the cache.
+func (c *Cache) Add(key []byte, value []byte) {
+	keyStr := string(key)
+	if _, ok := c.items[keyStr]; ok {
+		panic("slru: replacement of existing keys is not supported")
```
ckarenz: This edge case could be handled, but I was unsure what correct behavior would be. I chose to be explicit about not handling it.  If this were to pass unchecked, the cache would leak capacity with each duplicate add.

ckarenz:
___
*[slru/slru.go, line 88 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/16#-LcZqoP3SkxPIyIHGH1Z-r1-88:-LcigflUA55iiKW2454C:bsgijs9) ([raw file](https://github.com/dgraph-io/ristretto/blob/14e3b8e84bf43d276459ac6e89ba1f1baf351f52/slru/slru.go#L88)):*
<details><summary><i>Previously, ckarenz (Colin Arenz) wrote…</i></summary><blockquote>

This edge case could be handled, but I was unsure what correct behavior would be. I chose to be explicit about not handling it.  If this were to pass unchecked, the cache would leak capacity with each duplicate add.
</blockquote></details>

Quick followup: TinyLFU will not hit this case, so leaving it unhandled is reasonable.

ckarenz:
The implementation which uses this has drifted.  I'll draft a new PR with the full implementation.

PR #17, opened 2019-04-24
Port TinyLFU eviction policy from Caffeine

ckarenz:
The implementation included in this change is non-concurrent. It allows
injection of admission policies, but provides none yet. Further, storage
of key/value is left as a separate exercise; this tracks only whether a
key should be cached or evicted. Some small improvements have been made
over Caffeine to reduce allocation of list nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/17)

### DIFF
--- a/tinylfu/list_test.go
+++ b/tinylfu/list_test.go
@@ -1,0 +1,180 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tinylfu
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPushFront(t *testing.T) {
+	t.Run("PushFront", func(t *testing.T) {
+		l := newList()
+		assert.Nil(t, l.Front())
+
+		var elements []*element
+		for i := 0; i < 3; i++ {
+			e := &element{Value: uint64(i)}
+			elements = append([]*element{e}, elements...)
+			l.PushFront(e)
+
+			assert.Equal(t, e, l.Front())
+			checkList(t, l, elements)
+		}
+	})
+
+	t.Run("ChangeLists", func(t *testing.T) {
+		l1, l2 := newList(), newList()
+		e := &element{Value: 42}
+		l1.PushFront(e)
+		l2.PushFront(e)
+		checkList(t, l1, nil)
+		checkList(t, l2, []*element{e})
+	})
+}
+
+func TestPushBack(t *testing.T) {
+	t.Run("PushFront", func(t *testing.T) {
+		l := newList()
+		assert.Nil(t, l.Back())
+
+		var elements []*element
+		for i := 0; i < 3; i++ {
+			e := &element{Value: uint64(i)}
+			elements = append(elements, e)
+			l.PushBack(e)
+
+			assert.Equal(t, e, l.Back())
+			checkList(t, l, elements)
+		}
+	})
+
+	t.Run("ChangeLists", func(t *testing.T) {
+		l1, l2 := newList(), newList()
+		e := &element{Value: 42}
+		l1.PushBack(e)
+		l2.PushBack(e)
+		checkList(t, l1, nil)
+		checkList(t, l2, []*element{e})
+	})
+}
+
+func TestRemove(t *testing.T) {
+	t.Run("Uninitialized", func(t *testing.T) {
+		e := &element{}
+		assert.NotPanics(t, func() { e.Remove() })
+	})
+
+	t.Run("SingleElement", func(t *testing.T) {
+		e := &element{}
+		l := newList()
+		l.PushFront(e)
+		checkList(t, l, []*element{e})
+		e.Remove()
+		checkList(t, l, nil)
+	})
+
+	// Test removal of the head, middle, and tail.
+	for i := 0; i < 3; i++ {
+		l := newList()
+		var elements []*element
+		var remove *element
+
+		for ei := 0; ei < 3; ei++ {
+			e := &element{Value: uint64(ei)}
+			l.PushBack(e)
+			if ei == i {
+				remove = e
+			} else {
+				elements = append(elements, e)
+			}
+		}
+
+		t.Run(fmt.Sprintf("Remove%dOf3", i), func(t *testing.T) {
+			remove.Remove()
+			assert.Nil(t, remove.prev)
+			assert.Nil(t, remove.next)
+			assert.Nil(t, remove.list)
+			checkList(t, l, elements)
+		})
+	}
+}
+
+func TestMoveToFront(t *testing.T) {
+	t.Run("Uninitialized", func(t *testing.T) {
+		e := &element{}
+		assert.Panics(t, func() { e.MoveToFront() })
+	})
+
+	t.Run("SingleElement", func(t *testing.T) {
+		e := &element{}
+		l := newList()
+		l.PushFront(e)
+		e.MoveToFront()
+		checkList(t, l, []*element{e})
+	})
+
+	// Test removal of the head, middle, and tail.
+	for i := 0; i < 3; i++ {
+		l := newList()
+		var elements []*element
+		var move *element
+
+		for ei := 0; ei < 3; ei++ {
+			e := &element{Value: uint64(ei)}
+			l.PushBack(e)
+			if ei == i {
+				move = e
+				elements = append([]*element{e}, elements...)
+			} else {
+				elements = append(elements, e)
+			}
+		}
+
+		t.Run(fmt.Sprintf("Move%dOf3", i), func(t *testing.T) {
+			move.MoveToFront()
+			checkList(t, l, elements)
+		})
+	}
+}
+
+func checkList(t *testing.T, l *list, elements []*element) {
+	t.Helper()
+
+	root := &l.root
+	if !assert.Equal(t, len(elements), l.Len(), "list length") {
+		return
+	}
+
+	// Special case: empty list
+	if len(elements) == 0 {
+		if root.next != nil && root.next != root || root.prev != nil && root.prev != root {
+			t.Errorf("l.root.next = %p, l.root.prev = %p; both should both be nil or %p", l.root.next, l.root.prev, root)
+		}
+		return
+	}
+
+	for i, e := range elements {
+		assert.Equal(t, l, e.List())
+
+		if i > 0 {
+			assert.Equal(t, elements[i-1], e.prev, "internal prev pointer")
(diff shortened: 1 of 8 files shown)


ckarenz:
___
*[tinylfu/option.go, line 28 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCyo9l4J-X69Cyr6NT:-LdCyo9l4J-X69Cyr6NU:by1t2l2) ([raw file](https://github.com/dgraph-io/ristretto/blob/f5d1ba9170c9832efd2bd4b6f3eb8c699410ef2f/tinylfu/option.go#L28)):*
> ```Go
> 	RecordHit()
> 	RecordEviction()
> }
> ```

I chose to keep this relatively faithful to a similar recorder found in Caffeine.  More powerful instrumentation than pure counts would be possible by including the key as a parameter to each function.

ckarenz:
___
*[tinylfu/option.go, line 50 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCzEWgDq08zVqOx1uU:-LdCzEWhCKIe0FLBcjdS:b-8qcxnn) ([raw file](https://github.com/dgraph-io/ristretto/blob/f5d1ba9170c9832efd2bd4b6f3eb8c699410ef2f/tinylfu/option.go#L50)):*
> ```Go
> //   Protected = 1000 * (0.99 * 0.8) = 792
> //   Probation = 1000 - 792 - 10     = 198
> func WithSegmentation(main, protected float64) Option {
> ```

This is written to match Caffeine.  I tried a number of variations with trade-offs (convenience, clarity, predictability).  In the end I chose to stick with what works and provide a long-form explanation.  This version adds overrides for special cases to prevent 0-length segments.

ckarenz:
___
*[tinylfu/tinylfu.go, line 85 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdD-N4-2UXaI5p2Uuwb:-LdD-9YR5X6eRdSpS0Mm:bf37824) ([raw file](https://github.com/dgraph-io/ristretto/blob/f5d1ba9170c9832efd2bd4b6f3eb8c699410ef2f/tinylfu/tinylfu.go#L85)):*
> ```Go
> 
> // onMiss adds the entry to the admission window, evicting if necessary.
> func (p *Policy) onMiss(key uint64) {
> ```

This function contains the main change from Caffeine.  It guarantees we will not allocate list nodes after reaching full capacity.

ckarenz:
___
*[tinylfu/tinylfu.go, line 85 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdD-N4-2UXaI5p2Uuwb:-LdD5gWO2BljV3qeUvnW:bvuc16v) ([raw file](https://github.com/dgraph-io/ristretto/blob/f5d1ba9170c9832efd2bd4b6f3eb8c699410ef2f/tinylfu/tinylfu.go#L85)):*
<details><summary><i>Previously, ben-manes (Ben Manes) wrote…</i></summary><blockquote>

I think you looked at the simulator's version. In that case a miss is assumed to be implicit load, where no value is needed for a trace. I agree it doesn't try to optimize reusing the list node. That code's structure is mostly for simplicity, readability, and ease of hacking other variations (e.g. different admittor types, forking).

The production-grade library, though, would is more like a Map and the code structure is a lot different due to all of the other crazy concerns (like concurrency, HashDoS, user API).

The simulator's hopefully communicates intent, but definitely don't feel beholden to its structure. I'm sure you'll have more idiomatic styles and optimizations that wouldn't have made sense in that analysis code.
</blockquote></details>

I presumed we could compose the cache by slotting in algorithms similar to the simulator.  We can iterate toward a more tightly integrated approach if necessary, though I worry about testability.

ckarenz:
___
*[tinylfu/option.go, line 28 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCyo9l4J-X69Cyr6NT:-LdD7PQ22VnfEmNSfs3j:bz51ue5) ([raw file](https://github.com/dgraph-io/ristretto/blob/f5d1ba9170c9832efd2bd4b6f3eb8c699410ef2f/tinylfu/option.go#L28)):*
<details><summary><i>Previously, ben-manes (Ben Manes) wrote…</i></summary><blockquote>

Oh, in Caffeine the StatsCounter is for the application stats and not used directly for internal algorithms. We record stats independently, e.g. for count-min or hill climbing.
</blockquote></details>

That's the case here too.  If not injected via the option, stats recording no-ops.

tinylfu/tinylfu.go
```diff
@@ -0,0 +1,122 @@
...
+
+	p := &Policy{
+		data:      make(map[uint64]*element),
+		window:    newList(),
+		probation: newList(),
+		protected: newList(),
+		capacity:  capacity,
+	}
+
+	WithSegmentation(0.99, 0.8)(p)
```
vedhavyas: Since the ratios might depend on the workloads, why not take these values through Option and fallback to these if not given ?

tinylfu/list_test.go
```diff
@@ -0,0 +1,180 @@
...
+		e := &element{Value: 42}
+		l1.PushFront(e)
+		l2.PushFront(e)
+		checkList(t, l1, nil)
+		checkList(t, l2, []*element{e})
+	})
+}
+
+func TestPushBack(t *testing.T) {
+	t.Run("PushFront", func(t *testing.T) {
```
sourabh1024: PushBack ?

tinylfu/tinylfu.go
```diff
@@ -0,0 +1,122 @@
...
+	capacity     int
+	maxWindow    int
+	maxProtected int
+}
+
+// New creates a new TinyLFU cache.
+func New(capacity int, opts ...Option) *Policy {
+	// Consistent behavior relies on capacity for one element in each segment.
+	if capacity < 3 {
+		panic("tinylfu: capacity must be positive")
```
sourabh1024: the message should be more descriptive. capacity must be greater than 3.

tinylfu/tinylfu.go
```diff
@@ -0,0 +1,122 @@
+// Package tinylfu is an implementation of the W-TinyLFU caching algorithm.
+// See details at http://arxiv.org/abs/1512.00727
+package tinylfu
+
+// Policy implements a windowed TinyLFU eviction policy. It is not safe for concurrent access.
+type Policy struct {
+	data     map[uint64]*element
+	admittor AdmissionPolicy
```
mangalaman93: we can potentially move the configuration to another struct and define all the functions (With*) on that config instead. Check this out https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

tinylfu/tinylfu.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Record updates the policy when an entry is accessed.
+func (p *Policy) Record(key uint64) {
+	if p.admittor != nil {
+		p.admittor.Record(key)
+	}
+
+	node, ok := p.data[key]
+	if !ok {
+		if p.stats != nil {
```
mangalaman93: Instead of checking this everywhere, we can set the stats to a default StatsRecorder, which could be something meaningful or just could be a noop. Same for AdmissionPolicy.

ckarenz:
___
*[tinylfu/list.go, line 11 at r2](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCyCla2MLADbZbfg8x-r2-11:-LddmuDj5dnZBjlIGQ8O:b-bs73c9) ([raw file](https://github.com/dgraph-io/ristretto/blob/b9a8e20e10e338bb2963407afe028b8c3a5a376d/tinylfu/list.go#L11)):*
<details><summary><i>Previously, mangalaman93 (Aman Mangal) wrote…</i></summary><blockquote>

Shouldn't this be a pointer to the element (*element)? Any specific reason to keep it as a value.
</blockquote></details>

This isn't a real node, but a sentinel which must always exist and whose address must not change while the list contains any nodes.  I can add expand the comment to clarify.

ckarenz:
___
*[tinylfu/tinylfu.go, line 8 at r3](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCyClbHSlc8ZdsOuFJ-r3-8:-LdeMskK9MpQruPYqgqT:b-imf33a) ([raw file](https://github.com/dgraph-io/ristretto/blob/bb973dddd3249e3952d702436a737cd906966b2e/tinylfu/tinylfu.go#L8)):*
<details><summary><i>Previously, mangalaman93 (Aman Mangal) wrote…</i></summary><blockquote>

we can potentially move the configuration to another struct and define all the functions (With*) on that config instead. Check this out https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
</blockquote></details>

Apologies, but I don't understand the request.  The scheme implemented here is exactly Dave Cheney's proposed functional options.  Can you provide an example of what you mean?

ckarenz:
___
*[tinylfu/tinylfu.go, line 24 at r2](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCyClbHSlc8ZdsOuFJ-r2-24:-LdVm7o34fY4gg89QlYl:b-896fix) ([raw file](https://github.com/dgraph-io/ristretto/blob/b9a8e20e10e338bb2963407afe028b8c3a5a376d/tinylfu/tinylfu.go#L24)):*
<details><summary><i>Previously, sourabh1024 (Sourabh Suman) wrote…</i></summary><blockquote>

the message should be more descriptive. capacity must be greater than 3.
</blockquote></details>

Done.

ckarenz:
___
*[tinylfu/list_test.go, line 41 at r2](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCyCla2MLADbZbfg8y-r2-41:-LdVmDWW3rgEljpUwdSC:b-896fix) ([raw file](https://github.com/dgraph-io/ristretto/blob/b9a8e20e10e338bb2963407afe028b8c3a5a376d/tinylfu/list_test.go#L41)):*
<details><summary><i>Previously, sourabh1024 (Sourabh Suman) wrote…</i></summary><blockquote>

PushBack ?
</blockquote></details>

Done.

ckarenz:
___
*[tinylfu/tinylfu.go, line 55 at r3](https://reviewable.io/reviews/dgraph-io/ristretto/17#-LdCyClbHSlc8ZdsOuFJ-r3-55:-LdgTUljB2_3s07JRpEQ:b-896fix) ([raw file](https://github.com/dgraph-io/ristretto/blob/bb973dddd3249e3952d702436a737cd906966b2e/tinylfu/tinylfu.go#L55)):*
<details><summary><i>Previously, mangalaman93 (Aman Mangal) wrote…</i></summary><blockquote>

Instead of checking this everywhere, we can set the stats to a default StatsRecorder, which could be something meaningful or just could be a noop. Same for AdmissionPolicy.
</blockquote></details>

Done.

PR #18, opened 2019-05-21
BP-Wrapper striped ring buffer package

karlmcguire:
Simple striped, lossy (?) ring buffer implementation used in the "batching" process described in [BP-Wrapper](http://www.ece.eng.wayne.edu/~sjiang/pubs/papers/ding-09-BP-Wrapper.pdf) Section III part A. 

The tests aren't as extensive as I'd like, but I want to make sure that batching will work before trying for full coverage.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/18)


ring/ring.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Buffers stores a slice of buffers (stripes) and evenly distributes element
+// pushing between them.
+//
+// This implements the "batching" process described in the BP-Wrapper paper
+// (section III part A).
+type Buffers struct {
+	rows []*Buffer
+	mask uint64
+	busy uint64
```
golangcibot: `busy` is unused (from `structcheck`)

ring/README.md
```diff
@@ -0,0 +1,14 @@
...
+## benchmarks
+
+Benchmarks are ran on an Intel i7-8700K (3.7GHz 6-core).
+
+| Stripes | Size | Parallel | Ops/Second  | Allocs/Op |
+|:-------:|:----:|:--------:|:------------|:----------|
+| 1       | 128  |          | 131,140,000 | 0 |
+| 16      | 128  |          | 97,960,000  | 0 |
+| 1       | 128  | ✔        | 32,440,000  | 0 |
+| 16      | 128  | ✔        | 47,720,000  | 0 |
```
ben-manes: It is concerning that this goes down as the number of stripes increases. Ideally it would scale linearly up to the number of cpus. Ideal N cpus => N stripes, then some extra due to contention (thread migration, hash collisions). I suspect that your striping algorithm could be refined.

ring/ring.go
```diff
@@ -0,0 +1,100 @@
...
+		// so we can push the data to the Consumer
+		if atomic.CompareAndSwapUint64(&b.busy, 0, 1) {
+			// push buffer contents to the consumer
+			b.cons.Push(append(b.data[:0:0], b.data...))
+			// since the old elements are cleared out, place the new element
+			// at the front of the buffer
+			b.data[0] = element
+			// reset buffer head and unlock the busy mutex
+			atomic.StoreUint64(&b.head, 0)
+			atomic.StoreUint64(&b.busy, 0)
```
ben-manes: These two atomics are not composable, so you might have a few empty slots in the beginning due to a race (e.g. context switch). The consumer would need to be tolerant of absent values.

ring/ring.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	for i := range buffers.rows {
+		buffers.rows[i] = newBuffer(size, consumer)
+	}
+
+	return buffers
+}
+
+func (b *Buffers) id() uint64 {
```
ben-manes: I think that multiple threads could read the same value, stampede on each other, and use the same buffers. Ideally you would have a goroutine id (thread id) to hash with, and probe to a low contention mapping.

ring/README.md
karlmcguire: The parallel benchmarks are ran using an undefined number of goroutines. The Go team refuses to expose thread information, [despite using it in the standard library](https://github.com/golang/go/issues/18590#issuecomment-271661760). I'd much prefer to use the thread id for picking stripes, so right now I'm just randomly picking until I can find a more efficient way.

ring/ring.go
karlmcguire: You're right, goroutines are unfortunately anonymous by design. I'm considering using [runtime.LockOSThread()](https://golang.org/pkg/runtime/#LockOSThread) as a contrived way of doing this.

ben-manes: I think that multiple threads could read the same value, stampede on each other, and use the same buffers. Ideally you would have a goroutine id (thread id) to hash with, and probe to a low contention mapping.

ring/README.md
ben-manes: There is this cgo hack:
https://github.com/davecheney/junk/tree/master/id

In other systems one often has a fallback chain, testing if a low-level hack is supported and degrading until a language approved version works. Then you can fix when optimizations stop working without breaking applications.

karlmcguire:
It doesn't need to be random. It doesn't necessarily need to be an atomic int either, it can use a goroutine-local int to iterate through stripes. I will check the performance of the round-robin approach...

karlmcguire:
You were correct - round robin approach appears to be faster (and simpler).

PR #19, opened 2019-05-30
store: hash map package

karlmcguire:
We need a hash map for O(1) lookup of entries. This package will decouple this portion of Ristretto from everything else and allow exploration of hash map optimizations. 

Currently the default implementation is using `sync.Map` because of its great read performance. I've also included a regular `map` wrapped in a `sync.Mutex` for future testing. Depending on the data distribution, users may want to choose write-optimized or read-optimized hash maps. Another factor in the choice of hash map is key size, which could affect the hashing function performance.

Overall, this is a small abstraction but gives us the option of more engineering later down the road.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/19)


karlmcguire:
> > _[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/19)_ status: all files reviewed, 3 unresolved discussions (waiting on @karlmcguire)
> 
> _[store/store.go, line 27 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/19#-LgArkPg1n6LHf-Xs7TN:-LgArkPg1n6LHf-Xs7TO:bt36qjb) ([raw file](https://github.com/dgraph-io/ristretto/blob/0f92df4aec6868c27e2d75bdc80a5304fc3b6d5e/store/store.go#L27)):_
> 
> > ```go
> > type Map interface {
> > 	// Get returns the value associated with the key parameter.
> > 	Get(string) interface{}
> > ```
> 
> This should be `Get(string) uint64` as per our earlier discussion?

We've resolved this elsewhere, but for future reference:

The "user" of `store` is going to be the root `ristretto` package. All memory management should be done within the `store` package and return pointers (`interface{} or []byte` - whatever we decide) from its functions. This abstracts out the storage (and GC) part of the cache, and lets `ristretto` handle the higher level cache things.

PR #20, opened 2019-05-31
bench: benchmarking framework with CSV output

karlmcguire:
This is a simple benchmarking framework that outputs CSV for easy input into Google Sheets allowing us to make graphs like this:

Get-Samekey Performance (Parallel)

Right now it only measures speed, but I will add hit ratio measurement in the near future. I'm roughly basing it off of @mangalaman93's [cachebench](https://github.com/dgraph-io/benchmarks/tree/master/cachebench), but trying to rely on external libraries as little as possible. I expect to support Zipf and a variety of trace formats suggested by @ben-manes (ARC, LIRS, and Corda).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/20)


bench/cache_basemutex.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"sync"
+
+	"github.com/golang/groupcache/lru"
+)
+
+type BenchBaseMutex struct {
+	sync.Mutex
+	cache *lru.Cache
+	stats Stats
```
golangcibot: `stats` is unused (from `structcheck`)

bench/bench_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	// write csv data
+	records = append([][]string{Labels()}, records...)
+	// create file for writing
+	file, err := os.OpenFile(*PATH, os.O_WRONLY|os.O_CREATE, 0666)
+	if err != nil {
+		log.Panic(err)
+	}
+	// clear contents
+	file.Truncate(0)
```
golangcibot: Error return value of `file.Truncate` is not checked (from `errcheck`)

bench/bench_test.go
```diff
@@ -0,0 +1,89 @@
...
+		n := time.Now()
+		// get testing.BenchMarkResult
+		result := testing.Benchmark(benchmark.Bencher(benchmark))
+		// append to logs
+		logs = append(logs, &Log{benchmark, NewResult(result)})
+		log.Printf("\t ... %v\n", time.Since(n))
+	}
+
+	// save CSV to disk
+	save(logs)
```
golangcibot: Error return value of `save` is not checked (from `errcheck`)

bench/cache_bigcache.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (c *BenchBigCache) Set(key string, value interface{}) {
+	if err := c.cache.Set(key, value.([]byte)); err != nil {
+		log.Panic(err)
+	}
+}
+
+func (c *BenchBigCache) Del(key string) {
+	c.cache.Delete(key)
```
golangcibot: Error return value of `c.cache.Delete` is not checked (from `errcheck`)

bench/cache_ristretto.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+func (c *BenchRistretto) Get(key string) interface{} {
+	value, _ := c.cache.Get([]byte(key))
+	return value
+}
+
+func (c *BenchRistretto) Set(key string, value interface{}) {
+	c.cache.Set([]byte(key), value.([]byte))
```
golangcibot: Error return value of `c.cache.Set` is not checked (from `errcheck`)

ben-manes:
A few notes for future reference if helpful when this evolves to add more features:
- Concurrency:
    - Like `cachebench`, it will be desirable to include Zipf distribution in addition to a single-key (both in Caffeine's). Both are great at showing how the cache handles contention for hot entries.
    - Other distributions will probably not be very helpful for a concurrency benchmark
- Hit rates
   - You may prefer to add ristretto to the Java's trace simulator via go-java bindings
   - The simulator uses an actor model (channel + goroutine) and broadcasts to each actor a batch of events (1 message = 1000 events). Then it sends a shutdown poison pill and to receive back each policy's statistics. Each policy is wrapped by an actor to manage its interactions, so that the policy implementations use a simple, synchronous api (`record(long event)`, `finish()`, `getStats()`).  
  - The trace readers and report formats are decoupled, making it easy to plug in new types
  - The configuration to externalized to a file and the `Config` object is given to each policy to parse out its sections.
   - This is all relatively simple, multi-threaded, and fast. Somehow most I've seen are brittle and slow. Should be trivial to structure similarly in Golang if desired.
- Other perf benchmarks
  - You will probably want to benchmark the sequential cost of insert + eviction. That's just inserting a new key based on a counter. Some implementations can be surprisingly slow, e.g. Java's Ehcache 3 did O(n) scans to construct its sample for LRU, which degraded quickly (300 evictions per second at 10M entries). Basic sanity benchmarks, like you've been writing, are useful and oddly few do them.

PR #23, opened 2019-07-23
cost aware eviction

karlmcguire:
Converted everything to consider cost (size) of items when doing sampled LFU eviction. We now can return multiple victims when adding one item, depending on the sizes.

This is a complicated problem and in the future I will look at other solutions (GreedyDual, subset sum minimization, etc.) but for now I've kept it simple. Eviction is worst case O(n) because if there's 1,000 1kb items in the 1mb cache and you add a 1mb item, all old items need to be evicted. 

Other changes:

* Added doorkeeper (bloom filter) in front of the count-min sketch for TinyLFU
* Added `MaxBytes` and `MaxItems` fields to Cache `Config` struct
* Removed unused policies
* Updated tests
* Added `OnEvict` callback as requested
* Flattened out sub directory packages into root `ristretto` package

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/23)


dgryski:
Instead of Greedy Dual, there's also GD-Wheel ( http://2013.ladisworkshop.org/papers/ladis2013_submission_34.pdf ) and CAMP ( http://dblab.usc.edu/Users/papers/CAMPTR.pdf )

karlmcguire:
Thanks @dgryski, I'll take a look.

cache.go
```diff
@@ -17,79 +17,90 @@
...
 	return c.data.Get(key)
 }
 
-func (c *Cache) Set(key string, value interface{}) {
-	// attempt to add the key to the admission/eviction policy
-	if victim, added := c.policy.Add(key); added {
-		// delete the victim if there was an eviction
-		if victim != "" {
-			c.data.Del(key)
+func (c *Cache) Set(key string, val interface{}, cost uint64) ([]string, bool) {
```
ben-manes: Would it make more sense to evaluate through an attached function?

In Caffeine, we offer a `Weigher` function that computes an `int32` given the (key, value) pair. The is a nice property for computations, e.g. loading on a cache miss atomically. An explicit value upfront isn't known, so passing through is less useful in a concurrent setting.

The total `maximumWeight` of the cache is an `int64`. This avoids integer overflow and those using byte-sizes we expect them to scale appropriately. I'd make sure you have tests to if the sum exceeds `MaxUint64`.

Because we know upfront whether the cache is weight- (cost) or size-based, we code generate optimized versions of the cache entry. If the cache is size-bounded (weigher => 1) then we drop the weight field for a slimmer object. All of the logic remains the same, it is simply ignored by the entry factory when allocating the instance and its accessor returns the constant value.

karlmcguire: > Would it make more sense to evaluate through an attached function?

I think that's a great idea (reminds me of [Hyperbolic Caching](https://dl.acm.org/citation.cfm?id=3154690.3154738)) with the lazily evaluated priority function using item metadata as parameters. I'll have to ask @manishrjain about the origins of the cost parameter, and whether or not it would be advantageous to replace it with a function pointer. Right now the goal is just playing nicely with Badger. 

> The is a nice property for computations, e.g. loading on a cache miss atomically. An explicit value upfront isn't known, so passing through is less useful in a concurrent setting.

This is a good point. We don't have an atomic load-on-miss method right now, and if we added one the passed function would have to return both the value and cost. Do you think there's a benefit to decoupling the cost retrieval function from the value retrieval function?

> Because we know upfront whether the cache is weight- (cost) or size-based, we code generate optimized versions of the cache entry. If the cache is size-bounded (weigher => 1) then we drop the weight field for a slimmer object. All of the logic remains the same, it is simply ignored by the entry factory when allocating the instance and its accessor returns the constant value.

I'd like to add this optimization in the future. Perhaps rather than generating the code I'll just use another Policy implementation with less overhead in the case of a size-bounded configuration. I can reuse the shared logic between Policies.

ben-manes: > with the lazily evaluated priority function using item metadata as parameters.

Note that the function is evaluated on insertion or update, not on eviction, obviously because you need to track the totals. The entry's size may change and that must be done via a cache write for proper accounting, e.g. if it grew and forces an eviction. In a concurrent setting the race conditions has to be handled, such as when shifting entries between LRU lists, which is harder when the hash-table and policy are decoupled under different locks.

> the passed function would have to return both the value and cost. Do you think there's a benefit to decoupling the cost retrieval function from the value retrieval function?

Yes, I think it is simpler since the cost evaluation will be uniform (e.g. `len(data)`). In a simple load-on-miss case it might be simple if coupled, you could also have other complex cases to support such as cache refresh (`reload(key, oldValue)`), bulk load optimizations (`getAll(keys)` => `loadAll(keys)` for missing entries), and ad hoc computations (`compute(key, (key, oldValue) -> newValue)`). You would otherwise litter the evaluation everywhere, which is more error prone.

Another example is an async cache, where the value is a future, the entry's size is zero while in-flight (not eligible for eviction). When the future completes then a callback updates the cache, causing the entry's size to be calculated and an eviction to be triggered. The user-facing apis are much simpler / uniform / orthogonal with a function, imho.

PR #29, opened 2019-07-30
bench - show hit ratios and throughput

karlmcguire:
Updated bench to work with recent changes. Also made some aesthetic changes to `stats.csv` generation for clarity. 

Example `stats.csv` output:

```
name       , label        , go,  mop/s,  ns/op, ac, byt, hits    , misses  ,   ratio
ristretto  , hits-uniform ,  4, ------, ------, --, ---, 00239699, 00770402,  23.73%
ristretto  , hits-zipf    ,  4, ------, ------, --, ---, 02125603, 00884498,  70.62%
ristretto  , get-same     ,  4,  19.24,     51, 00, 008, --------, --------, -------
ristretto  , get-zipf     ,  4,  14.07,     71, 00, 008, --------, --------, -------
ristretto  , set-get      ,  4,   6.17,    162, 01, 032, --------, --------, -------
ristretto  , set-same     ,  4,   4.07,    245, 03, 056, --------, --------, -------
ristretto  , set-zipf     ,  4,   1.28,    782, 05, 089, --------, --------, -------
ristretto  , set-get-zipf ,  4,   1.43,    697, 03, 063, --------, --------, -------
```

Some areas are dashed-out to be less confusing. For example, hit ratio tracking messes with the throughput numbers. I think it's better to keep things separated.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/29)

### DIFF
--- a/bench/bench.go
+++ b/bench/bench.go
@@ -225,14 +225,14 @@ func Labels() []string {
 	return []string{
 		"name       ",
 		"label        ",
-		"gr",
-		"ops   ",
+		"go",
+		" mop/s",
+		" ns/op",
 		"ac",
 		"byt",
 		"hits    ",
 		"misses  ",
 		"  ratio ",
-		"  ns/op   ",
 	}
 }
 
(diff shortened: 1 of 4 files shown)


ben-manes:
In the throughput benchmarks, this has a 100% hit rate? Otherwise that would skew the numbers to be less realistic, how so depending on test implementation. You might separate into different reports (hit rates, throughput) just to keep it clear as two different functions.

PR #33, opened 2019-08-02
Overhaul the stats collection

manishrjain:
Make stats collection part of cache and pass it into policy and sampled LFU, so we can track lot more metrics. In particular:

- Number of keys added, updated, evicted.
- Cost of keys added and evicted.
- Sets dropped during channel push and rejected by policy.
- Gets dropped during channel push and kept.

Instead of using decorator pattern, this is using if statements to avoid doing atomic increments if the user is not interested in collecting metrics. I see no impact on performance between these two patterns.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/33)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -108,10 +108,12 @@ func (c *Cache) Get(key interface{}) (interface{}, bool) {
 	hash := z.KeyToHash(key)
 	c.buffer.Push(hash)
 	val, ok := c.data.Get(hash)
-	if ok {
-		c.stats.Add(hit, 1)
-	} else {
-		c.stats.Add(miss, 1)
+	if c.stats != nil {
+		if ok {
+			c.stats.Add(hit, 1)
+		} else {
+			c.stats.Add(miss, 1)
+		}
 	}
 	return val, ok
 }
--- a/cache.go
+++ b/cache.go
@@ -233,7 +233,9 @@ func (p *metrics) Add(t metricType, hash, delta uint64) {
 		return
 	}
 	valp := p.all[t]
-	idx := (hash % 25) * 10 // Leave space between two counters.
+	// Avoid false sharing by padding at least 64 bytes of space between two
+	// atomic counters which would be incremented.
+	idx := (hash % 25) * 10
 	atomic.AddUint64(valp[idx], delta)
 }
 
--- a/cache.go
+++ b/cache.go
@@ -108,12 +108,10 @@ func (c *Cache) Get(key interface{}) (interface{}, bool) {
 	hash := z.KeyToHash(key)
 	c.buffer.Push(hash)
 	val, ok := c.data.Get(hash)
-	if c.stats != nil {
-		if ok {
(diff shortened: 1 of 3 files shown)


cache.go
```diff
@@ -134,6 +148,119 @@ func (c *Cache) Del(key interface{}) {
...
+	if p == nil {
+		return ""
+	}
+	var buf bytes.Buffer
+	for i := 0; i < doNotUse; i++ {
+		t := metricType(i)
+		fmt.Fprintf(&buf, "%s: %d ", stringFor(t), p.Get(t))
+	}
+	fmt.Fprintf(&buf, "gets-total: %d", p.Get(hit)+p.Get(miss))
+	return string(buf.Bytes())
```
golangcibot: S1030: should use buf.String() instead of string(buf.Bytes()) (from `gosimple`)

cache.go
```diff
@@ -134,6 +148,120 @@ func (c *Cache) Del(key interface{}) {
...
+		return ""
+	}
+	var buf bytes.Buffer
+	for i := 0; i < doNotUse; i++ {
+		t := metricType(i)
+		fmt.Fprintf(&buf, "%s: %d ", stringFor(t), p.Get(t))
+	}
+	fmt.Fprintf(&buf, "gets-total: %d ", p.Get(hit)+p.Get(miss))
+	fmt.Fprintf(&buf, "hit-ratio: %.2f", p.Ratio())
+	return string(buf.Bytes())
```
golangcibot: S1030: should use buf.String() instead of string(buf.Bytes()) (from `gosimple`)

ben-manes:
Note that in a deployed application, you will want to optimize the counters as these are hot (contention on incrementing a hit). The simplest scheme is to use an array of counters, padded by the cache line size to avoid false sharing. Then hash into any to increment. You then summarize this array to return the total count for that metric.

For now this is a non-issue, but will be a bottleneck if enabled in a benchmark or application.

ben-manes:
do you have padding? It is likely worse because of false sharing of neighboring counters, rather than contention on that one counter. You can space them apart.

I assume a 64-byte cache line, though that may no longer be accurate on the latest hardware.

ben-manes:
http://www.drdobbs.com/parallel/eliminate-false-sharing/217500206 is a good walk through of false sharing of a counter array

ben-manes:
that is odd, but glad it is now at an acceptable overhead. I don't know why being at the of the lines boundary would impact it. I have heard that now loads are 2 cache lines wide in newest hardware, so one must pad by 128 bytes, but that seemed excessive and I didn't fully grok the explanation.

ben-manes:
You may also want to review other places (like read buffers) for false sharing. I padded all of my variables to be safe, though that was probably excessive, by being lazy to focus on code. I never got back to validating to only pad where needed, as it is tricky to determine those hotspots.

Issue #34, opened 2019-08-02
`bench` fails to build


h8liu:
```
# github.com/dgraph-io/ristretto/bench
src/github.com/dgraph-io/ristretto/bench/bench.go:325:10: undefined: ristretto.PolicyLog
```

Found by gopkgs.io

karlmcguire:
Thanks for the heads up. We just overhauled the hit ratio tracking in Ristretto so I'm going to have to decouple `bench` and add my own tracking. Give me a few hours.

manishrjain:
Sorry for the breakage. Did the change late night and was more focused on the performance and coverage. Thanks Karl for taking care of it.

karlmcguire:
Fixed in #35.

PR #35, opened 2019-08-14
bench hit ratio updates

karlmcguire:
- ARC and LIRS trace file parsing
- cache configs matching @dgryski's cachetest, hit ratios matching as well
- move hit ratio tests to single thread/goroutine
- decoupled Policy stats collection from `ristretto` package (fixing #34)
- improved Clairvoyant/Optimal Policy because we're running massive ARC traces and it needs to be fast (it should be O(n lg n))
- documentation of `sim` package, a start to comprehensive documentation

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/35)


bench/cache.go
```diff
@@ -33,25 +34,104 @@ type Cache interface {
...
+type BenchOptimal struct {
+	capacity uint64
+	hits     map[string]uint64
+	access   []string
+}
+
+func NewBenchOptimal(capacity int, track bool) Cache {
+	return &BenchOptimal{
+		capacity: uint64(capacity),
+		hits:     make(map[string]uint64, 0),
```
golangcibot: S1019: should use make(map[string]uint64) instead (from `gosimple`)

PR #37, opened 2019-08-27
Replace AESHash with MemHash

ashish-goswami:
Currently we are using AESHash(aeshash in runtime) which uses
aesenc instruction. This can crash restritto if it is running
on a hardware not having this instruction. Go runtime, at init
decides to use memhash if this instruction is not available.
memhash is independent of hardware. memhash on every call,
checks if aeshash can be calculated(based on the availabilty of
aesenc intruction), if yes it uses aeshash. If not, it uses normal flow(memhash).
This commit changes AESHash to MemHash.

Fixes: https://github.com/dgraph-io/badger/issues/980

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/37)

### DIFF
--- a/z/rtutil.go
+++ b/z/rtutil.go
@@ -40,21 +40,23 @@ type stringStruct struct {
 }
 
 //go:noescape
-//go:linkname aeshash runtime.aeshash
-func aeshash(p unsafe.Pointer, h, s uintptr) uintptr
+//go:linkname memhash runtime.memhash
+func memhash(p unsafe.Pointer, h, s uintptr) uintptr
 
-// AESHash is the hash function used by map, it utilizes available hardware instructions.
+// MemHash is the hash function used by map, it utilizes available hardware instructions(behaves as
+// aeshash if aes instruction is available).
 // NOTE: The hash seed changes for every process. So, this cannot be used as a persistent hash.
-func AESHash(data []byte) uint64 {
+func MemHash(data []byte) uint64 {
 	ss := (*stringStruct)(unsafe.Pointer(&data))
-	return uint64(aeshash(ss.str, 0, uintptr(ss.len)))
+	return uint64(memhash(ss.str, 0, uintptr(ss.len)))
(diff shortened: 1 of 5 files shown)


z/rtutil.go
```diff
@@ -40,21 +40,23 @@ type stringStruct struct {
 }
 
 //go:noescape
-//go:linkname aeshash runtime.aeshash
-func aeshash(p unsafe.Pointer, h, s uintptr) uintptr
+//go:linkname memhash runtime.memhash
+func memhash(p unsafe.Pointer, h, s uintptr) uintptr
 
-// AESHash is the hash function used by map, it utilizes available hardware instructions.
+// MemHash is the hash function used by map, it utilizes available hardware instructions(behaves as
```
jarifibrahim: `go maps` instead of `map` would make more sense.

z/rtutil.go
```diff
@@ -40,21 +40,23 @@ type stringStruct struct {
...
 // NOTE: The hash seed changes for every process. So, this cannot be used as a persistent hash.
-func AESHash(data []byte) uint64 {
+func MemHash(data []byte) uint64 {
 	ss := (*stringStruct)(unsafe.Pointer(&data))
-	return uint64(aeshash(ss.str, 0, uintptr(ss.len)))
+	return uint64(memhash(ss.str, 0, uintptr(ss.len)))
 }
 
-// AESHashString is the hash function used by map, it utilizes available hardware instructions.
+// MemHashString is the hash function used by map, it utilizes available hardware instructions
```
jarifibrahim: `used by go maps`

ashish-goswami: Done.

z/rtutil.go
ashish-goswami: Done.

ashish-goswami:
GODEBUG=cpu.aes=off, disables aeshash.

PR #51, opened 2019-09-22
fix onevict

### LINKED ISSUE #49: Possible bug when processing items
I was reading through the code to see how you implemented "ristretto" and I came up to this line: https://github.com/dgraph-io/ristretto/blob/master/cache.go#L201.
```go
c.onEvict(item.key, item.val, item.cost)
```
I might be wrong, because I don't yet fully understand the code involved, but shouldn't it be:
```go
c.onEvict(victim.key, victim.val, victim.cost)
```

karlmcguire:
fixes #49

I also added a test to make sure this bug won't pop up again.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/51)

### DIFF
--- a/cache_test.go
+++ b/cache_test.go
@@ -105,7 +105,7 @@ func newRatioTest(cache TestCache) func(t *testing.T) {
 
 func TestCacheOnEvict(t *testing.T) {
 	mu := &sync.Mutex{}
-	evictions := make(map[uint64]int, 0)
+	evictions := make(map[uint64]int)
 	cache, err := NewCache(&Config{
 		NumCounters: 1000,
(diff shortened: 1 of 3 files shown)


cache_test.go
```diff
@@ -104,13 +104,16 @@ func newRatioTest(cache TestCache) func(t *testing.T) {
 }
 
 func TestCacheOnEvict(t *testing.T) {
-	evictions := 0
+	mu := &sync.Mutex{}
+	evictions := make(map[uint64]int, 0)
```
golangcibot: S1019: should use make(map[uint64]int) instead (from `gosimple`)

PR #53, opened 2019-09-22
Add a way to stop goroutines

hnakamur:
First of all, thanks for a wonderful library!

This pull request adds a way to stop goroutines as stated in TODOs.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/53)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -47,8 +47,6 @@ type Cache struct {
 	// setBuf is a buffer allowing us to batch/drop Sets during times of high
 	// contention
 	setBuf chan *item
-	// closeCh is a channel to stop goroutines which is running processingItems.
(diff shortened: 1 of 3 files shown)


cenkalti:
Using a context here is not a proper fit in my opinion. It forces the caller to save passed context in a variable.
Using a stop channel that will get closed when `Close()` is called is simpler.

karlmcguire:
I'm going to take a close look at this. In the past I've tried doing a simple `closeCh` on the `processItems`, but constantly selecting over the two channels (rather than ranging over the one) really hurt performance.

hnakamur:
Thanks! :tada:

PR #61, opened 2019-09-24
Stop goroutines when closing cache

tuanpavn:
Thank for an amazing library. I am really looking forward to using the first release version.

This pull request creates a `processCh` channel that is identical to `setBuf` channel.
When cache is closing, `setBuf` is set to `nil` to avoid adding new items into buffer and `processCh` is closed to release goroutines.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/61)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"sync"
 	"sync/atomic"
 
 	"github.com/dgraph-io/ristretto/z"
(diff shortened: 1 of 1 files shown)


tuanpavn:
> Thanks for your contribution. Don't like the idea of using a lock around a channel. In fact, you can just close setBuf. A user is expected to not call Set after having closed the cache. If they still do that, they'll get a panic, which makes sense, they should fix their code.
> 
> > Reviewed 1 of 1 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/61#-:-LpdCUQ67Wj4GgzUHlJH:b-mivykr)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jarifibrahim)

Even user is expected to not call `Set`, it is possible that ongoing call has not exited `Set` function, it will cause `panic` when send to closed channel.  And lock used in `Set` is read lock, it won't hurt performance.

PR #71, opened 2019-09-26
refactor(valName): local variable name same with global struct

OhBonsai:
Hi,  `policy.go:L389` use `item` as variable which is conflict with `cache.go:L96`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/71)


karlmcguire:
I took the opportunity to remove some unnecessary lines in that function. We're currently not using `lruPolicy` but may want to in the future, so I added it to the testing suite as well. Thanks for catching the bad variable name!

Issue #68, opened 2019-09-25
Add mechanism to clear entire cache.


martinmr:
I am adding ristretto to Dgraph but I am seeing a few issues that are caused because the cache is not being cleared after a DropAll operation.

I tried fixing this by recreating the entire cache every time that a DropAll op is processed, but this leads to out of memory issues. Ideally, ristretto has an API to efficiently clear the entire contents of the cache.

I'll take a look at the code and see if I can come up with something myself, but most likely I will need help from someone more familiar with the internals of ristretto.

PR #73, opened 2019-09-27
basic Clear operation

karlmcguire:
This should help in #68. The only caveat at the moment is `setBuf` isn't cleared (not sure how to best do that safely).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/73)

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -45,6 +45,8 @@ type policy interface {
 	Cap() int64
 	// Optionally, set stats object to track how policy is performing.
 	CollectMetrics(stats *metrics)
+	// Clear zeroes out all counters and clears hashmaps.
+	Clear()
 }
 
 func newPolicy(numCounters, maxCost int64) policy {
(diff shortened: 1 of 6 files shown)


policy.go
```diff
@@ -251,6 +260,13 @@ func (p *sampledLFU) updateIfHas(key uint64, cost int64) (updated bool) {
 	return false
 }
 
+func (p *sampledLFU) clear() {
+	p.used = 0
+	for k, _ := range p.keyCosts {
```
golangcibot: S1005: should omit value from range; this loop is equivalent to `for k := range ...` (from `gosimple`)

store.go
```diff
@@ -118,3 +105,11 @@ func (m *lockedMap) Del(key uint64) {
 	defer m.Unlock()
 	delete(m.data, key)
 }
+
+func (m *lockedMap) Clear() {
+	m.Lock()
+	defer m.Unlock()
+	for k, _ := range m.data {
```
golangcibot: S1005: should omit value from range; this loop is equivalent to `for k := range ...` (from `gosimple`)

PR #75, opened 2019-09-27
immediate Updates and lazy cost evaluation

karlmcguire:
This PR addresses the Set/Del race condition by putting Dels on the same internal buffer, as well as immediately updating Set values and eventually updating the corresponding cost.

Costs can also now be calculated using the `config.Coster` function.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/75)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -51,6 +51,8 @@ type Cache struct {
 	// Each key will be hashed using the provided function. If keyToHash value
 	// is not set, the default keyToHash function is used.
 	keyToHash func(interface{}) uint64
+	// coster calculates cost from a value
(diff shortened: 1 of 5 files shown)


ben-manes:
Is there a reason you prefer `cost` and `Coster` versus the Java terms (`weight` and `Weigher`)?(When refer you referred to the entry's size in your article, "a heavy item could displace many lightweight items", you used weight as the metaphor)

karlmcguire:
I was just going to change it to `Cost`, but I like the idea of using `Weigher` for the reason you mentioned... I will check with @manishrjain.

Issue #59, opened 2019-09-24
more structured access to metrics.


tcolgate:
Metrics returns a private type that doesn't show it's methods in the GoDoc. Ideally I'd like to be able to expose the metrics to a prometheus collector. It looks like I could possibly parse some string formatted output, but, ideally, I'd be able to get a current set of structuted mwtrics from the call.

PR #84, opened 2019-10-02
export Metrics in a cleaner struct

karlmcguire:
#59

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/84)


tcolgate:
This looks good.

tcolgate:
> There should be a way to only retrieve a specific metric instead of all the metrics every time.

From my PoV, and I suspect a lot of users of this, I'll be sticking a prom collector on the front of this and exposing all the metrics on every scrape. That seems like it would be the more common use case.

karlmcguire:
> There should be a way to only retrieve a specific metric instead of all the metrics every time.

I agree, and we can do that internally. But for the exported function, a "snapshot" with all the struct fields seems good enough.

PR #89, opened 2019-10-07
remove dependency testify

andersfylling:
I noticed that two dependencies can easily be removed by adjusting 7 lines of code. This also removes the redundant uint64 casting as by default, consts are untyped.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/89)


andersfylling:
I'm unable to replicate the failure of TestCacheProcessItems

karlmcguire:
@andersfylling Thanks for the PR. 

As for the test, it's not your fault. I'm pretty sure the `time.Sleep(time.Millisecond)` doesn't simply isn't long enough for the CI server so it "fails." I need to figure out a better mechanism for those time-sensitive tests (for now I'll probably just increase the sleep time).

Issue #92, opened 2019-10-21
Race in cache.clear


jarifibrahim:
The following race condition was seen in badger on PR https://github.com/dgraph-io/badger/pull/1066
```
------- Stdout: -------
=== RUN   TestDropAllWithPendingTxn
badger 2019/10/21 08:34:34 INFO: All 0 tables opened in 0s
badger 2019/10/21 08:34:35 DEBUG: Flushing memtable, mt.size=34867 size of flushChan: 0
badger 2019/10/21 08:34:35 DEBUG: Storing value log head: {Fid:0 Len:30 Offset:4831937}
badger 2019/10/21 08:34:35 DEBUG: Storing value log head: {Fid:1 Len:30 Offset:4026620}
badger 2019/10/21 08:34:35 DEBUG: Flushing memtable, mt.size=34837 size of flushChan: 0
badger 2019/10/21 08:34:35 DEBUG: Storing value log head: {Fid:2 Len:30 Offset:3221294}
badger 2019/10/21 08:34:35 DEBUG: Flushing memtable, mt.size=34825 size of flushChan: 0
badger 2019/10/21 08:34:36 DEBUG: Storing value log head: {Fid:3 Len:30 Offset:2415968}
badger 2019/10/21 08:34:36 DEBUG: Flushing memtable, mt.size=34777 size of flushChan: 0
badger 2019/10/21 08:34:36 DEBUG: Storing value log head: {Fid:4 Len:30 Offset:1610642}
badger 2019/10/21 08:34:36 DEBUG: Flushing memtable, mt.size=34757 size of flushChan: 0
badger 2019/10/21 08:34:36 DEBUG: Storing value log head: {Fid:5 Len:30 Offset:805316}
badger 2019/10/21 08:34:36 DEBUG: Flushing memtable, mt.size=34753 size of flushChan: 0
badger 2019/10/21 08:34:36 INFO: Got compaction priority: {level:0 score:1.2 dropPrefix:[]}
badger 2019/10/21 08:34:36 INFO: Running for level: 0
badger 2019/10/21 08:34:36 DEBUG: LOG Compact. Added 3534 keys. Skipped 0 keys. Iteration took: 47.099272ms
badger 2019/10/21 08:34:36 DEBUG: Discard stats: map[]
badger 2019/10/21 08:34:36 INFO: LOG Compact 0->1, del 6 tables, add 1 tables, took 64.409113ms
[...]

PR #93, opened 2019-10-23
stats Clear race condition fix

karlmcguire:
Along with a small fix, I added `TestCacheMetricsClear` to simulate what I perceived as the issue in #92, which is:

**Goroutine 1:** Continuously calling cache.Get (incrementing metrics counters atomically)
**Goroutine 2:** Calls cache.Clear (calls collectMetrics() which creates a new metrics object and replaces the pointer in cache instance)

The simplest solution I could think of is reserving the collectMetrics() function for initialization, which performs well, but for the cache.Clear operation we should use the same metrics object and atomically zero all counters. This will allow us to keep the performance benefits of using a single atomic operation for metrics.Add calls and not having to use a mutex or channel.

The downside to this solution is the possibility of metric inaccuracies because of the iterative nature of the metrics.Clear operation. Because metrics.Clear is not a single atomic operation, parallel calls to metrics.Add will continue to mutate the metrics object while it's being zeroed.

I do not foresee this downside causing any issues in the future because it would only occur in situations where cache.Clear is being called in parallel with other cache operations, so the user has presumably given up the expectation of fully accurate metrics and would be unable to measure the accuracy.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/93)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -239,8 +239,8 @@ func (c *Cache) Clear() {
 	c.policy.Clear()
 	c.store.Clear()
 	// only reset metrics if they're enabled
-	if c.stats != nil {
-		c.stats.Clear()
+	if c.Metrics != nil {
+		c.Metrics.Clear()
 	}
 	// restart processItems goroutine
 	go c.processItems()
@@ -466,7 +466,7 @@ func (p *Metrics) Clear() {
 		}
 	}
 }
-  
+
 func (p *Metrics) String() string {
 	if p == nil {
 		return ""
--- a/cache.go
+++ b/cache.go
@@ -238,7 +238,7 @@ func (c *Cache) Clear() {
 	c.store.Clear()
 	// only reset metrics if they're enabled
 	if c.stats != nil {
-		c.collectMetrics()
+		c.stats.Clear()
 	}
 	// restart processItems goroutine
 	go c.processItems()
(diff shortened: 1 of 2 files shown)


ben-manes:
Why even clear metrics? Many reporting systems work better with monotonically incrementing counts, which defers the rollups (e.g. 5/10/15m) to their time-series database, which can then calculate the delta from the raw stats. If clear / rollups / etc are needed client-side then this can be done on-top, e.g. by baselining on a snapshot to diff against. Then only handling overflow is necessary.

karlmcguire:
> Why even clear metrics?

That's actually a good question. I will bring it up with Badger team and see.

karlmcguire:
@ben-manes I assume on overflow all metrics need to be reset?

ben-manes:
Once it overflows a 64-bit counter it is probably not worth worrying too much about. Instead when making the snapshot we simply convert a negative value to the maximum. 2^63 is quite large so it wasn't worth worrying and over complicating it. (see [ConcurrentStatsCounter](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/stats/ConcurrentStatsCounter.java) if interested)

For reference, @manishrjain should remember Borgmon which assumed monatomic counters.

jarifibrahim:
@karlmcguire I'm seeing the same race condition with this PR also https://teamcity.dgraph.io/viewLog.html?buildId=29750&buildTypeId=Badger_UnitTests .

jarifibrahim:
The fix works fine in badger https://github.com/dgraph-io/badger/pull/1099

Issue #43, opened 2019-09-17
Support TTL and reloadData


tuanpavn:
Please add TTL and loadData function into config so it can auto reload an item if it is expired.

barkanido:
Ttl eviction is highly valuable!

manishrjain:
We can do TTL based expiry. But load on TTL seems like a cumbersome design.

barkanido:
Agreed, just invalidate items due to TTL and let the client decide.

songboyu:
hello, could I ask when would you plan to add the ttl feature? I think it is necessary

MarcSky:
Yes we need ttl

MarcSky:
For example, I want than 1 key living 10second and after reload

karlmcguire:
Will be working on this over the next few days.

caofengl:
> Will be working on this over the next few days.
May I ask when the master can join the ttl function?

momilo:
If I may join the discussion... It might be worth considering adding automated purging of expired items. 

Ristretto currently seems to be focusing on a scenario where there is a huge amount of data, only a portion of which can be stored in pre-defined size of cache, and it is stored forever (i.e. MaxCost is reached quickly; some additional keys are tracked - NumCounters).

However, many use-cases require a TTL-driven cache, whereby you have to / are happy to cache _everything_  as long as ... e.g. it is max 5 minutes old (e.g. for operations which tend to always be resolved within 1-2 minutes). Hence you would expect the MaxCost to be large (just in case ever needed) but the average cache size would be minimal. A go routine checking expiry times and deleting entries would be needed - and highly appreciated.

I would be happy to help with coding this. Currently it seems like karlmcguire and caofengl are already on it, but feel free to reach out to me if some extra help is needed.

barkanido:
I think this was the original intent. wasn't it?

momilo:
My understanding was that the original intention was to merely add TTL as an eviction criterion, i.e. to remove expired items only when, and only to the extent that a space needs to be made for, a new item. But if a routine cleanup is already being considered then great.

barkanido:
@karlmcguire any update on this?

martinmr:
@barkanido This was just merged recently. See PR #122

martinmr:
Re-opening since only the TTL part is done. I am not sure what the status of the reload function is.

martinmr:
[...]

PR #104, opened 2019-11-15
added TTL

karlmcguire:
This is a simple solution for #43.

The only issue I can foresee is some of the random samples not picking up items with set TTLs and rather than evicting expired items we'd evict "indefinite" items within the sample. So, this TTL solution is "sampled TTL" and not true TTL.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/104)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -185,6 +185,9 @@ func (c *Cache) Get(key interface{}) (interface{}, bool) {
 // To dynamically evaluate the items cost using the Config.Coster function, set
 // the cost parameter to 0 and Coster will be ran when needed in order to find
 // the items true cost.
+//
+// TTL is the amount of time (in seconds) that the item will remain in the
+// cache. To make an item stay indefinitely, set TTL to -1.
 func (c *Cache) Set(key, value interface{}, cost, ttl int64) bool {
 	if c == nil || key == nil {
 		return false
--- a/cache.go
+++ b/cache.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"sync/atomic"
+	"time"
 
 	"github.com/dgraph-io/ristretto/z"
 )
@@ -189,6 +190,9 @@ func (c *Cache) Set(key, value interface{}, cost, ttl int64) bool {
 		return false
 	}
 	keyHash, conflictHash := c.keyToHash(key)
+	if ttl != -1 {
+		ttl = time.Now().Unix() + ttl
+	}
 	i := &item{
 		flag:     itemNew,
 		key:      keyHash,
--- a/cache.go
+++ b/cache.go
@@ -286,7 +286,7 @@ func (c *Cache) processItems() {
 					c.Metrics.add(costEvict, victim.key, uint64(victim.cost))
 				}
 			case itemUpdate:
-				c.policy.Update(i.key, i.cost, i.ttl)
+				c.policy.Update(i.key, i.cost)
 			case itemDelete:
 				c.policy.Del(i.key)
 				c.store.Del(i.key, i.conflict)
--- a/cache.go
+++ b/cache.go
@@ -120,6 +120,7 @@ type item struct {
 	conflict uint64
 	value    interface{}
 	cost     int64
+	ttl      int64
 }
 
 // NewCache returns a new Cache instance and any configuration errors, if any.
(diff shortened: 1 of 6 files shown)


policy.go
```diff
@@ -233,18 +264,23 @@ func (p *defaultPolicy) Close() {
 	close(p.itemsCh)
 }
 
+type metadata struct {
```
golangcibot: type `metadata` is unused (from `unused`)

policy.go
```diff
@@ -233,18 +264,23 @@ func (p *defaultPolicy) Close() {
...
+	cost int64
```
golangcibot: `cost` is unused (from `structcheck`)

policy.go
```diff
@@ -233,18 +264,23 @@ func (p *defaultPolicy) Close() {
...
+	ttl  int64
```
golangcibot: `ttl` is unused (from `structcheck`)

cache.go
```diff
@@ -183,17 +185,24 @@ func (c *Cache) Get(key interface{}) (interface{}, bool) {
 // To dynamically evaluate the items cost using the Config.Coster function, set
 // the cost parameter to 0 and Coster will be ran when needed in order to find
 // the items true cost.
-func (c *Cache) Set(key, value interface{}, cost int64) bool {
+//
+// TTL is the amount of time (in seconds) that the item will remain in the
+// cache. To make an item stay indefinitely, set TTL to -1.
+func (c *Cache) Set(key, value interface{}, cost, ttl int64) bool {
```
erikdubbelboer: This would be the moment to start tagging releases. Because you haven't tagged a release before everyone is using `master` and everyone's build will break at soon as this is merged.

karlmcguire: Agreed

ccampbell:
Just randomly stumbled upon this PR. I am obviously not super familiar with this code so take these comments with a grain of salt, but I am curious why you chose to implement TTL the way you did vs. doing something like this (pseudo code):

```go
func set(key, value, ttl) {
    exp = time.Now() + ttl
    store(key, wrapped{value, exp})
}

func get(key) value {
    wrapped = retrieve(key)
    if time.Now() > wrapped.exp {
        forceEvict(key)
        return nil
    }

    return wrapped.value
}
```
Obviously that is oversimplified, but this would guarantee that expired data can never be returned.

I could be wrong but in your implementation it looks like expired items are only purged when a new `cache.Set()` call occurs via `policy.Add()`. So if you have data that was stored with a ttl and has expired, and you never add new items to cache, then the get call for that content will return stale data until you add something new to your cache. Not saying that is likely, but it could happen.

---

Another advantage of storing the expiration along with the value is it could be used to automatically prevent cache stampeding or could be queried on its own. For example, you could have a separate `cache.GetStale()` method that could do something like

1. If the content is not in cache or is in cache and not expired return the item as usual
2. If the content is in cache, but expired return `nil` and immediately set the stale value back to cache with a future expiration time (one minute in the future, for example).

This would allow the first request that hits an expired item to go back to disk/database/wherever to repopulate the item with a fresh value while ensuring that all other requests to the content in the meantime get the stale value from cache preventing a big spike in traffic to regenerate the cache if a bunch of requests all are made for it at the same time.

Anyway, just some thoughts. I know this library is very performance focused, so I imagine there may be some performance considerations I am missing here.

BradErz:
@manishrjain Random stranger here :wave: hope you don't mind me butting in.

What about implementing the ttl as a variadic function?

```go
func (c *Cache) Set(key, value interface{}, cost, opts ...SetOption) bool {
....
}
```

And the option would be set something like:
```go
cache.Set("key", "value", 0, cache.WithTTL(ttl time.Duration))
```

To be honest i would also be tempted to do it for the cost parameter as well? That way only the required parameters are actually required but the optional ones can just assume defaults or be overridden by the client. 

I'm not familiar with the dgraph code base so not sure if you are already using the pattern somewhere else?

ben-manes:
@manishrjain fyi - the evolution of that map of buckets idea, called a [hierarchical timing wheel](https://www.confluent.io/blog/apache-kafka-purgatory-hierarchical-timing-wheels/), is implemented in Caffeine.

martinmr:
Closing in favor of https://github.com/dgraph-io/ristretto/pull/122

PR #113, opened 2019-12-20
delete item immediately

balajijinnah:
Delete was happening concurrently in a separate goroutine. This will make the delete happen immediately But the pitfall is that if there is a set in the channel it'll bring the item again. So, we're pushing to the channel as well to maintain the order. 
Signed-off-by: balaji <balaji@dgraph.io>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/113)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -216,7 +216,10 @@ func (c *Cache) Del(key interface{}) {
 		return
 	}
 	keyHash, conflictHash := c.keyToHash(key)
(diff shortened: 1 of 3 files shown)


jarifibrahim:
This PR needs a description.

balajijinnah:
@jarifibrahim updated

PR #119, opened 2020-01-11
Support nil *Cache values in Clear and Close

damz:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/119)


damz:
Unfortunately CI is broken in this repository, so I had to fix a test that seems to have been broken since at least f823dc4a5031a0404648c7e004c65d37bdf3f38d.

cache.go
```diff
@@ -225,8 +225,15 @@ func (c *Cache) Del(key interface{}) {
 
 // Close stops all goroutines and closes all channels.
 func (c *Cache) Close() {
+	if c == nil {
+		return
+	}
 	// block until processItems goroutine is returned
-	c.stop <- struct{}{}
```
jarifibrahim: @damz If I understand correctly, we should block here so that if `processItems()` is processing an item we wait for it to complete. 

https://github.com/dgraph-io/ristretto/blob/83508260cb49a2c3261c2774c991870fd18b5a1b/cache.go#L283-L287

Can you explain why this change was necessary?

damz: This is to make sure that that `Close` is idempotent, but that does indeed break the expectation that `processItems` is on its way out before the policy is closed. I pushed a new version with a different implementation.

PR #122, opened 2020-01-16
Sets with TTL

martinmr:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/122)


cache_test.go
```diff
@@ -524,3 +589,9 @@ func TestCacheMetricsClear(t *testing.T) {
 	c.Metrics = nil
 	c.Metrics.Clear()
 }
+
+func TestMain(m *testing.M) {
```
golangcibot: SA3000: TestMain should call os.Exit to set exit code (from `staticcheck`)

cache_test.go
```diff
@@ -524,3 +590,10 @@ func TestCacheMetricsClear(t *testing.T) {
...
+	// Set bucketSizeSecs to 1 to avoid waiting too much during the tests.
+	bucketDurationSecs = 1
+	m.Run()
+	os.Exit(0)
```
dominikh: You are supposed to exit with the return value of m.Run, so that the exit status of the test binary reflects whether tests have passed or failed.

But the whole TestMain function can probably be replaced with a `func init() { bucketDurationSecs = 1 }` in cache_test.go

PR #130, opened 2020-01-30
Remove key from policy after TTL eviction

### LINKED ISSUE #129: [PR #122] It's not possible to reuse a key after TTL eviction
Hello,

I've found an issue when playing with the new TTL feature.
It's not possible to reuse a key to update the cache with a new value.

The sample use case :
```
c, err := NewCache(...)

// Add a value with key 1 in the cache
c.SetWithTTL(1, 1, 1, 5*time.Second)
_, ok := c.Get(1) // return true

// some time later
// Add another value with key 1 in the cache
c.SetWithTTL(1, 1, 1, 5*time.Second)
_, ok := c.Get(1) // return false
```

I submit the following PR to correct this issue : #130

iotafr:
Fixes #129

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/130)

### DIFF
--- a/cache_test.go
+++ b/cache_test.go
@@ -330,6 +330,45 @@ func TestCacheSet(t *testing.T) {
 	require.False(t, c.Set(1, 1, 1))
 }
 
+func TestRecacheWithTTL(t *testing.T) {
+	c, err := NewCache(&Config{
+		NumCounters: 100,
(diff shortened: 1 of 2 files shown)


iotafr:
You're welcome !

PR #136, opened 2020-02-06
Use sync.Pool to use &item{}

choleraehyq:
```
benchmark                                    old ns/op     new ns/op     delta
BenchmarkConcurrentCacheSetNeverEvict-12     96.3          127           +31.88%

benchmark                                    old allocs     new allocs     delta
BenchmarkConcurrentCacheSetNeverEvict-12     3              2              -33.33%

benchmark                                    old bytes     new bytes     delta
BenchmarkConcurrentCacheSetNeverEvict-12     110           108           -1.82%
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/136)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


mangalaman93:
This PR is not relevant anymore

PR #139, opened 2020-02-19
Adding FastRandom func which can be called from other package

anandshah05:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/139)

### DIFF
--- a/z/rtutil.go
+++ b/z/rtutil.go
@@ -62,3 +62,8 @@ func MemHashString(str string) uint64 {
 // FastRand is a fast thread local random function.
 //go:linkname FastRand runtime.fastrand
 func FastRand() uint32
+
+// FastRandom is fast random number generator function
+func FastRandom(maxNo uint32) uint32 {
+	return FastRand() % maxNo
+}


z/rtutil.go
```diff
@@ -62,3 +62,8 @@ func MemHashString(str string) uint64 {
 // FastRand is a fast thread local random function.
 //go:linkname FastRand runtime.fastrand
 func FastRand() uint32
+
+// FastRandom is fast random number generator function
+func FastRandom(maxNo uint32) uint32 {
+	return FastRand() % maxNo
```
erikdubbelboer: Never use `%` for this: https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator

martinmr:
This is not used by Ristretto at the moment. If you find that function useful you are free to declare it in you own package(s) but the z package should contain utility functions for Ristretto and other Dgraph software (Badger uses the bloom filter).

PR #144, opened 2020-02-24
Protect against attacks.

### LINKED ISSUE #131: Protect against hash attacks.
An attacker can prevent new items from being accepted by the policy by artificially raising the frequency of the victim.

Caffeine code: https://github.com/ben-manes/caffeine/blob/8bdfb0d95db523d15c2560f05a228468ecdbe98d/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L752-L776

martinmr:
An attacker can prevent new items from being admitted to the cache by
artificially raising the frequency of the victims. This PR implements an
strategy used in Caffeine to occasionnaly let items that would otherwise
be rejected.

Fixes #131

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/144)

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -18,6 +18,7 @@ package ristretto
 
 import (
 	"math"
+	"math/rand"
 	"sync"
 
 	"github.com/dgraph-io/ristretto/z"
@@ -173,7 +174,11 @@ func (p *defaultPolicy) Add(key uint64, cost int64) ([]*item, bool) {
 		}
 
 		// If the incoming item isn't worth keeping in the policy, reject.
-		if incHits < minHits {
+		// However, a little bit of randomness is introduced to prevent against
+		// hash collision attacks, in which the score of the victims is artificially
+		// raised so that no new items are admitted into the cache.
+		// This strategy is implemented by Caffeine.
+		if incHits < minHits && !(rand.Int() % 127 == 0) {
 			p.metrics.add(rejectSets, key, 1)
 			return victims, false
 		}


policy.go
```diff
@@ -173,7 +174,11 @@ func (p *defaultPolicy) Add(key uint64, cost int64) ([]*item, bool) {
 		}
 
 		// If the incoming item isn't worth keeping in the policy, reject.
-		if incHits < minHits {
+		// However, a little bit of randomness is introduced to prevent against
+		// hash collision attacks, in which the score of the victims is artificially
+		// raised so that no new items are admitted into the cache.
+		// This strategy is implemented by Caffeine.
+		if incHits < minHits && !(rand.Int() % 127 == 0) {
```
ben-manes: You can replace the modulus `%` with a bitmask `&`. This was why 127 was chosen (2^n - 1) so that we could avoid the division at ~1% acceptance rate. The division vs mask was perhaps an over-optimization in this code path, though!

ben-manes:
Due to using random eviction, you currently are not impacted by this attack. Caffeine is because LRU is a deterministic algorithm. However I think having some protection is good regardless since it is cheap and documented if you refactor away from random eviction someday.

manishrjain (commented):
@martinmr What's the status of this? Do we still need it?

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/144#-:-MHGtd7PE8wIrGcbTnu5:b6fcbf7)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jarifibrahim, @karlmcguire, @manishrjain, and @martinmr)

martinmr:
Closing this ticket since it's not a required feature.
cc: @jarifibrahim

PR #145, opened 2020-02-25
Change public API to use byte arrays for keys.

### LINKED ISSUE #107: KeyToHash possibly letting key interfaces escape to heap
As brought up in the performance Gopher channel, since we're using `interface{}` for keys, [this call to `keyToHash`](https://github.com/dgraph-io/ristretto/blob/master/cache.go#L166) is probably letting the interface escape to heap.

Solution: switch over to `string` or `[]byte` for public API calls.

martinmr:
Fixes #107

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/145)


varun06:
Any chance for this change to move forward. We are also seeing cases where Ristretto is using more memory then maxCost, possibly because of cache key(interface{}) escaping to heap.

PR #160, opened 2020-06-05
 ristretto/z: use hash/memhash for hashing

rogpeppe:
This is cleaner, but unfortunately slower.

Reusing the Hash value in KeyToHash may help there,
but wouldn't get back to the original speed, so maybe
this isn't a reasonable change.

```
name             old time/op    new time/op    delta
MemHash-4          12.4ns ± 7%    43.1ns ± 1%  +248.30%  (p=0.008 n=5+5)
MemHashString-4    19.6ns ± 7%    70.9ns ±20%  +262.00%  (p=0.008 n=5+5)

name             old speed      new speed      delta
MemHash-4        5.17GB/s ± 7%  1.48GB/s ± 0%   -71.31%  (p=0.008 n=5+5)
MemHashString-4  6.29GB/s ± 6%  1.75GB/s ±18%   -72.12%  (p=0.008 n=5+5)

name             old alloc/op   new alloc/op   delta
MemHash-4           0.00B          0.00B           ~     (all equal)
MemHashString-4     0.00B          0.00B           ~     (all equal)

name             old allocs/op  new allocs/op  delta
MemHash-4            0.00           0.00           ~     (all equal)
MemHashString-4      0.00           0.00           ~     (all equal)
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/160)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/dgraph-io/ristretto
 go 1.12
 
 require (
-	github.com/cespare/xxhash v1.1.0
 	github.com/dgryski/go-farm v0.0.0-20190423205320-6a90982ecee2
 	github.com/stretchr/testify v1.4.0
 )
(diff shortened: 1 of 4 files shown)


manishrjain:
Thanks for the PR, @rogpeppe . Yeah, the speeds don't look encouraging, though I'm wondering why that's the case. My understanding was that Go 1.14 exposed the same functions.

rogpeppe:
Yeah, me too. I did it in a rush over about 20 minutes before bed last night so didn't have time to investigate more. I'm interested why it's so much slower - I might take another look today.

Issue #167, opened 2020-06-29
Update returns false even when the item was updated


jarifibrahim:
The bug was seen in https://github.com/dgraph-io/badger/issues/1350 . 
See https://github.com/dgraph-io/badger/issues/1350#issuecomment-650796966 for a detailed explanation.

The following `TestUpdateBug` test fails on current master https://github.com/dgraph-io/ristretto/commit/a093fe661329ba68a99a9c3a63674992355f40b1
```diff
diff --git a/cache.go b/cache.go
index e48a48a..c5e19f0 100644
--- a/cache.go
+++ b/cache.go
@@ -29,7 +29,7 @@ import (
 	"github.com/dgraph-io/ristretto/z"
 )
 
-const (
+var (
 	// TODO: find the optimal value for this or make it configurable
 	setBufSize = 32 * 1024
 )
diff --git a/cache_test.go b/cache_test.go
index c1dc745..7c0bd9b 100644
--- a/cache_test.go
+++ b/cache_test.go
@@ -1,7 +1,9 @@
 package ristretto
 
 import (
+	"fmt"
 	"math/rand"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
@@ -614,3 +616,54 @@ func init() {
 	// Set bucketSizeSecs to 1 to avoid waiting too much during the tests.
 	bucketDurationSecs = 1
 }
+
+func TestUpdateBug(t *testing.T) {
+	originalSetBugSize := setBufSize
+	defer func() { setBufSize = originalSetBugSize }()
+	test := func() {
+		droppedMap := make(map[int]struct{})
+		lastEvictedSet := int64(-1)
+
+		var err error
+		handler := func(_ interface{}, value interface{}) {
+			v := value.(string)
+			lastEvictedSet, err = strconv.ParseInt(string(v), 10, 32)
+			require.NoError(t, err)
+			_, ok := droppedMap[int(lastEvictedSet)]
+			if ok {
+				panic(fmt.Sprintf("val = %+v was dropped but got evicted. Dropped items: %+v\n", lastEvictedSet, droppedMap))
+			}
+		}
+
+		// This is important.
+		setBufSize = 10
+
+		c, err := NewCache(&Config{
+			NumCounters: 100,
+			MaxCost:     10,
+			BufferItems: 64,
+			Metrics:     true,
+			OnEvict: func(_, _ uint64, value interface{}, _ int64) {
+				handler(nil, value)
+			},
+		})
[...]

PR #168, opened 2020-06-29
Improve handling of updated items

jarifibrahim:
When an item is updated, the existing item was updated in the `store` but we might return false.
Update happens here https://github.com/dgraph-io/ristretto/blob/a093fe661329ba68a99a9c3a63674992355f40b1/cache.go#L229-L231
and the false is returned here https://github.com/dgraph-io/ristretto/blob/a093fe661329ba68a99a9c3a63674992355f40b1/cache.go#L232-L239

With this PR, we will always return true for updates, even if the policy wasn't updated. 
See #167 for more details.

Fixes https://github.com/dgraph-io/ristretto/issues/167

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/168)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -29,7 +29,7 @@ import (
 	"github.com/dgraph-io/ristretto/z"
 )
 
-var (
+const (
 	// TODO: find the optimal value for this or make it configurable
 	setBufSize = 32 * 1024
 )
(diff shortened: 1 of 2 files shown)


cache.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/dgraph-io/ristretto/z"
 )
 
-const (
+var (
```
jarifibrahim, line 32: This was changed so that I can change the `setBufSize` in the tests.

PR #178, opened 2020-08-03
use defaults suggested in docs explicitly in code

costela:
Please see [this duscussion](https://discuss.dgraph.io/t/provide-defaults-for-numcounters-and-bufferitems/9194) for a rationale for this suggestion.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/178)


sadqx (approved):

cache.go
```diff
@@ -158,14 +161,18 @@ type Item struct {
...
-		return nil, errors.New("NumCounters can't be zero")
-	case config.MaxCost == 0:
+	if config.MaxCost == 0 {
 		return nil, errors.New("MaxCost can't be zero")
-	case config.BufferItems == 0:
-		return nil, errors.New("BufferItems can't be zero")
 	}
+
+	if config.NumCounters == 0 {
+		config.NumCounters = config.MaxCost * defaultNumCountersMult // sensible default
```
pandeyshubham25, line 169: Since MaxCost is the upper bound on size of cache in Bytes, it doesn't necessarily approximate to max number of keys we want in it. This likely leads to over-approximation for value of NumCounters which should be 10 times the number of keys in cache and not 10 times the size of cache.

cache_test.go
```diff
@@ -133,7 +133,7 @@ func TestNewCache(t *testing.T) {
 		MaxCost:     10,
 		BufferItems: 0,
 	})
-	require.Error(t, err)
```
pandeyshubham25, line 136: Is it possible to add more test cases pertinent to the changes in defaulting behavior? This is to make sure that any future changes to this behavior are rightly caught by the test case(s)

pandeyshubham25, line 136: Is it possible to add more test cases pertinent to the changes in defaulting behavior? This is to make sure that any future changes to this behavior are rightly caught by the test case(s)

pandeyshubham25 (commented):
@costela , thank you for suggesting the change and working on the improvement. I have added a few comments inline and would appreciate it if you can look at those.

cache.go
costela: True, but I assumed over-approximation would be less harmful then under-approximation. Isn't this the case?

cache_test.go
costela: I just tried making these tests a bit more explicit. PTAL

I don't think we can significantly improve coverage further here: we can't realistically check for all cases where  we _don't_ expect an error, so we'd bound to covering the error cases. WDYT?

PR #179, opened 2020-08-12
Expose cache item expiration

corinapurcarea:
I'm contributing to a multi layer cache library that propagates entries from lower caches to upper ones and in order to do this in an effective manner, i need the TTL of the entry stored in ristretto.

I oscillated between adding Expiration and a GetWithTTL which would also return the item in addition to TTL. If you think a GetWithTTL it's a better alternative, let me know and i'll replace Expiration.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/179)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -181,6 +181,35 @@ func (c *Cache) Get(key interface{}) (interface{}, bool) {
 	return value, ok
 }
 
+// Expiration returns the duration and a boolean representing whether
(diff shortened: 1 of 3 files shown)


varun06:
This will be helpful in our use case too. 👍

karlmcguire:
Thanks for the PR @corinapurcarea! Looks great, I will get the tests passing and merge.

PR #180, opened 2020-08-17
Make item public. Add a new onReject call for rejected items.

manishrjain:
- Making Item public makes the onEvict and onReject function calls more readable.
- Adding onReject allows us to tightly track every Set that happens, so we can avoid memory leaks in manually allocated memory.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/180)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -34,7 +34,7 @@ var (
 	setBufSize = 32 * 1024
 )
 
-type onEvictFunc func(uint64, uint64, interface{}, int64)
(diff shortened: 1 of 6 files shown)


jarifibrahim:
The `TestDropUpdates` works fine locally. It could be a Travis issue (the test has a time.Sleep) or a go 1.12 . 
@anurags92 @jarifibrahim to investigate it later.

PR #188, opened 2020-08-25
Add histogram.Mean() method

jarifibrahim:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/188)

### DIFF
--- a/z/histogram.go
+++ b/z/histogram.go
@@ -98,6 +98,9 @@ func (histogram *HistogramData) Update(value int64) {
 
 // Mean returns the mean value for the histogram.
 func (histogram *HistogramData) Mean() float64 {
+	if histogram.Count == 0 {
+		return 0
+	}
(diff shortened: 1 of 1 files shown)


jarifibrahim:
The teamcity build is failing because the `test.sh` script is missing on master.

z/histogram.go
```diff
@@ -96,6 +96,11 @@ func (histogram *HistogramData) Update(value int64) {
 	}
 }
 
+// Mean returns the mean value for the histogram.
+func (histogram *HistogramData) Mean() float64 {
+	return float64(histogram.Sum) / float64(histogram.Count)
```
anurags92, line 101: No chance of someone calling it when histogram.count is zero? Will it crash or exit gracefully?

jarifibrahim: Nice catch. Fixed it.

anurags92 (approved):

PR #189, opened 2020-08-27
update comments to correctly reflect counter space usage

neena:
When using ristretto in our application, we noticed that the memory used per counter was much higher than the documentation suggested. This PR updates the comments in the config struct to accurately represent the space per counter (and matches our observations). 

Sources:
https://github.com/neena/ristretto/blob/29b4dd7a077785696ba5422081b3c301d4c1e5f1/sketch.go#L43-L62
https://github.com/neena/ristretto/blob/8f368f2f2ab3a54cbe62fb9772cd75ce55e07802/z/bbloom.go#L51-L75

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/189)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -77,8 +77,11 @@ type Config struct {
 	// accuracy and subsequent hit ratios.
 	//
 	// For example, if you expect your cache to hold 1,000,000 items when full,
-	// NumCounters should be 10,000,000 (10x). Each counter takes up 4 bits, so
-	// keeping 10,000,000 counters would require 5MB of memory.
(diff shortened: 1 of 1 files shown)


neena:
Hi - any feedback on this? I'd love to move it out of my PR inbox (or make changes as needed)! Let me know 😄

neena:
@manishrjain @karlmcguire Would it be possible to get a review on this? This caused some confusion for us when rolling out a ristretto cache in production, and it would be nice to avoid this confusion for future users!

PR #194, opened 2020-09-22
Introduce Mmapped buffers and Merge Sort

manishrjain:
Buffers can now be mmapped as well as Calloc'd. This PR also copies over all the mmap files from Badger to allow mmap support in various platforms.

This PR also introduces Merge Sort to do sorting of the buffer using an extra temporary space costing half of the space as the original buffer, currently allocated on Calloc. We can't use quick sort. Each entry is variable length, so we can't just swap them in the buffer. Merge Sort allows us to iterate over them linearly, hence is a better fit.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/194)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,5 +6,7 @@ require (
 	github.com/cespare/xxhash v1.1.0
 	github.com/dgryski/go-farm v0.0.0-20190423205320-6a90982ecee2
 	github.com/dustin/go-humanize v1.0.0
+	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.4.0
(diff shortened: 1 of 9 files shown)


jarifibrahim (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/194#-:-MHqHS_HD99lS67TOZ1X:bnfp4nl)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @anurags92, @karlmcguire, and @martinmr)

PR #209, opened 2020-10-27
fix(btree): Increase buffer size if needed.

NamanJain8:
This PR makes the mmap used for `B+ tree` file growable. One important point to see here is that some of the nodes need to be re-read.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/209)

### DIFF
--- a/z/btree.go
+++ b/z/btree.go
@@ -36,6 +36,13 @@ type Tree struct {
 	nextPage uint64
 }
 
+// Release the memory allocated to tree.
+func (t *Tree) Release() {
+	if t != nil && t.mf != nil {
+		t.mf.Delete()
+	}
+}
+
 // NewTree returns a memory mapped B+ tree.
 func NewTree(mf *MmapFile) *Tree {
 	// Tell kernel that we'd be reading pages in random order, so don't do read ahead.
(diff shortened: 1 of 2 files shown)


z/btree.go
```diff
@@ -93,10 +105,12 @@ func (t *Tree) Set(k, v uint64) {
 		panic("Error setting zero or MaxUint64")
 	}
 	root := t.node(1)
-	t.set(root, k, v)
+	root = t.set(1, k, v)
```
jarifibrahim, line 108: Add a comment.

z/btree.go
```diff
@@ -93,10 +105,12 @@ func (t *Tree) Set(k, v uint64) {
 		panic("Error setting zero or MaxUint64")
 	}
 	root := t.node(1)
```
jarifibrahim, line 107: Do we still need this? I think we can remove it.

z/btree.go
```diff
@@ -113,10 +127,11 @@ func (t *Tree) Set(k, v uint64) {
 
 // For internal nodes, they contain <key, ptr>.
 // where all entries <= key are stored in the corresponding ptr.
-func (t *Tree) set(n node, k, v uint64) {
+func (t *Tree) set(offset uint64, k, v uint64) node {
+	n := t.node(offset)
 	if n.isLeaf() {
 		n.set(k, v)
```
jarifibrahim, line 133: return n.set

z/btree.go
```diff
@@ -113,10 +127,11 @@ func (t *Tree) Set(k, v uint64) {
...
-		return
+		return n
```
jarifibrahim, line 134: This can be combined with the previous statement.

z/btree.go
```diff
@@ -132,19 +147,23 @@ func (t *Tree) set(n node, k, v uint64) {
...
-
+	child = t.set(child.pageID(), k, v)
+	// Re-read n as the underlying buffer for tree might have changed during set.
+	n = t.node(offset)
 	if child.isFull() {
-		nn := t.split(child)
-
+		nn := t.split(child.pageID())
+		n = t.node(offset)
+		child = t.node(n.uint64(valOffset(idx)))
```
jarifibrahim, line 159: Add a comment.

jarifibrahim (approved):

PR #213, opened 2020-11-02
Fixmemtest

chewxy:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/213)


chewxy:
@mrjn I added code to do a comparison between stdlib 's calloc vs jemalloc. So now you can just run the `memtest` program to use the correct allocator.

PR #214, opened 2020-11-02
Fix infinite loop in allocator

ahsanbarkati:
If we create a new allocator starting with zero size, then addBufferWithMinSize() will run into an infinite loop, trying to achieve the required buffer size by doubling the pagesize (which is zero).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/214)

### DIFF
--- a/z/allocator.go
+++ b/z/allocator.go
@@ -115,9 +115,9 @@ func ReturnAllocator(a *Allocator) {
 func NewAllocator(sz int) *Allocator {
 	ref := atomic.AddUint64(&allocRef, 1)
 	// We should not allow a zero sized page because addBufferWithMinSize
-	// will run into an infinite loop tyring to double the pagesize.
+	// will run into an infinite loop trying to double the pagesize.
 	if sz == 0 {
-		sz = 1
+		sz = smallBufferSize
 	}
 	a := &Allocator{
 		pageSize: sz,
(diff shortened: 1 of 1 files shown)


z/allocator.go
```diff
@@ -114,6 +114,11 @@ func ReturnAllocator(a *Allocator) {
 // NewAllocator creates an allocator starting with the given size.
 func NewAllocator(sz int) *Allocator {
 	ref := atomic.AddUint64(&allocRef, 1)
+	// We should not allow a zero sized page because addBufferWithMinSize
+	// will run into an infinite loop tyring to double the pagesize.
+	if sz == 0 {
+		sz = 1
```
jarifibrahim, line 120: maybe use https://github.com/dgraph-io/ristretto/blob/4f21aeb8a0424f16df7843a276030ba8bd8fb53f/z/buffer.go#L72 ?

jarifibrahim (approved):

z/allocator.go
```diff
@@ -373,7 +378,6 @@ func (a *Allocator) Allocate(sz int) []byte {
 		}
 		cb = a.buffers[a.curBuf]
 	}
-
```
jarifibrahim, line 376: Please add the new line back.

PR #215, opened 2020-11-03
chore(buffer): add a parent directory field in z.Buffer

NamanJain8:
This PR adds a `dir` field inside z.Buffer. This field is the parent directory for the mmaped buffers created. If the field is not specified(i.e., it is ""), then it uses `/tmp` (default).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/215)

### DIFF
--- a/z/buffer.go
+++ b/z/buffer.go
@@ -74,11 +74,19 @@ const smallBufferSize = 64
 
 // NewBuffer is a helper utility, which creates a virtually unlimited Buffer in UseCalloc mode.
 func NewBuffer(sz int) *Buffer {
-	buf, err := NewBufferWith(sz, 256<<30, UseCalloc, "")
+	buf, err := NewBufferWithDir(sz, 256<<30, UseCalloc, "")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


z/buffer.go
```diff
@@ -107,7 +108,7 @@ func (b *Buffer) doMmap() error {
 // NewBufferWith would allocate a buffer of size sz upfront, with the total size of the buffer not
 // exceeding maxSz. Both sz and maxSz can be set to zero, in which case reasonable defaults would be
 // used. Buffer can't be used without initialization via NewBuffer.
-func NewBufferWith(sz, maxSz int, bufType BufferType) (*Buffer, error) {
+func NewBufferWith(sz, maxSz int, bufType BufferType, dir string) (*Buffer, error) {
```
jarifibrahim, line 111: `NewBufferWithDir(...)`

jarifibrahim (approved):

z/buffer.go
NamanJain8: Done.

PR #216, opened 2020-11-03
fix(buffer): Make padding internal to z.buffer

ahsanbarkati:
The caller of the API should not be aware of the padding. It should be handled internally.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/216)


jarifibrahim (approved):
Add a test for it.

PR #221, opened 2020-11-10
Use mremap on Linux instead of munmap+mmap

ajeetdsouza:
Linux exposes an `mremap` function that is not available on other OSes, that allows us to remap an address space, providing a more efficient way to increase the size of an mmap file. This PR allows Linux systems to use the `mremap` call, while other OSes continue to use `munmap` followed by `mmap`.

`sys` does not provide an `mremap` function, so this had to handwritten using a system call. Unfortunately, the functions provided by `sys` for `Mmap`/`Munmap` maintain a list of active mappings, and do not allow us to `Munmap` anything that is not present in the list. Since our handwritten `mremap` call cannot update that internal list, we must reimplement the `munmap` system call as well.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/221)

### DIFF
--- a/z/file_default.go
+++ b/z/file_default.go
@@ -1,0 +1,23 @@
+// +build !linux
+
+package z
+
+import "fmt"
+
+// Truncate would truncate the mmapped file to the given size. On Linux and
+// others, we could directly just truncate the underlying file, but in Windows,
+// we can't do that. So, unmap first, then truncate, then re-map.
+func (m *MmapFile) Truncate(maxSz int64) error {
+	if err := m.Sync(); err != nil {
+		return fmt.Errorf("while sync file: %s, error: %v\n", m.Fd.Name(), err)
+	}
+	if err := Munmap(m.Data); err != nil {
+		return fmt.Errorf("while munmap file: %s, error: %v\n", m.Fd.Name(), err)
+	}
+	if err := m.Fd.Truncate(maxSz); err != nil {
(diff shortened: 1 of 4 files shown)


z/file_linux.go
```diff
@@ -0,0 +1,21 @@
+package z
```
jarifibrahim, line 1: missing licence

z/file_default.go
```diff
@@ -0,0 +1,23 @@
+// +build !linux
```
jarifibrahim, line 1: missing licence

z/mmap_linux.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, fmt.Errorf("mremap size mismatch: requested: %d got: %d", size, mmapSize)
+	}
+
+	header.Data = mmapAddr
+	header.Cap = size
+	header.Len = size
+	return data, nil
+}
+
+func munmap(data []byte) error {
```
jarifibrahim, line 63: Add a comment about why it's needed.

z/mmap_linux.go
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2019 Dgraph Labs, Inc. and Contributors
```
jarifibrahim, line 2: 2020

jarifibrahim (approved):
Clean it up and it's good to merge

PR #236, opened 2020-12-07
fix(jemalloc): Fix the stats of jemalloc

ahsanbarkati:
The jemalloc stats are read by using the `mallctl()` call, this call doesn't return the latest stats unless we refresh the ctl cache. We can get the latest stats after calling an epoch mallctl. This PR fixes this issue.

**Note: Calling an epoch mallctl is expensive. An epoch mallctl takes up all the locks as taken by a malloc call. We should use this judiciously.**

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/236)

### DIFF
--- a/contrib/memtest/main.go
+++ b/contrib/memtest/main.go
@@ -33,6 +33,7 @@ import (
 	"unsafe"
 
 	"github.com/dgraph-io/ristretto/z"
+	"github.com/dustin/go-humanize"
 )
 
 type S struct {
@@ -109,8 +110,16 @@ func memory() {
 			counter++
 		}
 	}
-	fmt.Printf("[%d] Current Memory: %05.2f G. Increase? %v\n",
-		counter, float64(curMem)/float64(1<<30), increase)
+	var js z.MemStats
+	z.ReadMemStats(&js)
+
+	fmt.Printf("[%d] Current Memory: %s. Increase? %v, MemStats [Active: %s, Allocated: %s,"+
+		" Resident%s, Retained: %s]\n",
+		counter,
+		humanize.IBytes(uint64(z.NumAllocBytes())),
+		increase,
+		humanize.IBytes(js.Active), humanize.IBytes(js.Allocated),
+		humanize.IBytes(js.Resident), humanize.IBytes(js.Retained))
 }
 
 func viaLL() {
--- a/contrib/memtest/main.go
+++ b/contrib/memtest/main.go
@@ -115,7 +115,7 @@ func memory() {
 
 	fmt.Printf("[%d] Current Memory: %s. Increase? %v, MemStats [Active: %s, Allocated: %s,"+
 		" Resident: %s, Retained: %s]\n",
-		counter, humanize.IBytes(uint64(z.NumAllocBytes())), increase,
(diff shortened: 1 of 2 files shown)


contrib/memtest/main.go
```diff
@@ -109,8 +110,14 @@ func memory() {
 			counter++
 		}
 	}
-	fmt.Printf("[%d] Current Memory: %05.2f G. Increase? %v\n",
-		counter, float64(curMem)/float64(1<<30), increase)
+	var js z.MemStats
+	z.ReadMemStats(&js)
```
jarifibrahim, line 114: Do you need this in memtest?

jarifibrahim (approved):
Looks good to me.

contrib/memtest/main.go
ahsanbarkati: I was thinking that this can be something which we can use to verify that the stats are as expected.

manishrjain (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/236)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @ahsanbarkati, @jarifibrahim, @manishrjain, and @martinmr)

___
*[contrib/memtest/main.go, line 114 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/236#-MNxwcRoMP4gtmDY0Emc-r1-114:-MNy7_DV8BcDFge88np5:b-4i5s8z) ([raw file](https://github.com/dgraph-io/ristretto/blob/1f52196f94fae030734175791760602b59992bfa/contrib/memtest/main.go#L114)):*
<details><summary><i>Previously, ahsanbarkati (Ahsan Barkati) wrote…</i></summary><blockquote>

I was thinking that this can be something which we can use to verify that the stats are as expected.
</blockquote></details>

The memstats can be added on. But, our version of how much memory should be used should still be printed.
___
*[z/calloc_jemalloc.go, line 164 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/236#-MNy7iFICbMufThNdkAB:-MNy7iFICbMufThNdkAC:b-bes4po) ([raw file](https://github.com/dgraph-io/ristretto/blob/1f52196f94fae030734175791760602b59992bfa/z/calloc_jemalloc.go#L164)):*
> ```Go
> 	sz := unsafe.Sizeof(&epoch)
> 	C.je_mallctl(
> 		(C.CString)("epoch"),
> ```

Link to documentation about this?

ahsanbarkati:
___
*[contrib/memtest/main.go, line 114 at r1](https://reviewable.io/reviews/dgraph-io/ristretto/236#-MNxwcRoMP4gtmDY0Emc-r1-114:-MNyHGIQ2XXvmGymNdvi:bgyiei) ([raw file](https://github.com/dgraph-io/ristretto/blob/1f52196f94fae030734175791760602b59992bfa/contrib/memtest/main.go#L114)):*
<details><summary><i>Previously, manishrjain (Manish R Jain) wrote…</i></summary><blockquote>

The memstats can be added on. But, our version of how much memory should be used should still be printed.
</blockquote></details>

Yep, that is still printed.

PR #237, opened 2020-12-17
fix(b+tree): Do not double the size of btree

ahsanbarkati:
We should not double the size of b+tree because it is being used in xidmap which can grow big according to the data. Doubling will cause unnecessary memory usage.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/237)

### DIFF
--- a/z/btree.go
+++ b/z/btree.go
@@ -114,9 +114,14 @@ func (t *Tree) newNode(bit uint64) node {
 		pageId = t.nextPage
 		t.nextPage++
 		offset := int(pageId) * pageSize
-		// Double the size if current buffer is insufficient.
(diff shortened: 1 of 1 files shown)


NamanJain8 (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/237#-:-MOls1-LDMhQy9IKYy-c:bg7igkv)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jarifibrahim, @manishrjain, and @martinmr)

NamanJain8 (approved):

PR #243, opened 2021-01-12
Add feature to be able to list cached keys

Darkweak:
Hi, for a side-project (a cache system called [Souin](https://github.com/darkweak/Souin)), I'm planning to expose the cached keys through a REST API
Then I need the feature to list them. Due to the hashed keys I couldn't retrieve corresponding keys, then I choose to store keys before hashing.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/243)

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -78,6 +78,9 @@ type Cache struct {
 	// Metrics contains a running log of important statistics like hits, misses,
 	// and dropped items.
 	Metrics *Metrics
+	// keySaver contains the saved keys as string representation only if
+	// EnableKeySaver is set in Config, nil otherwise
+	keySaver *ClearKey
 }
 
 // Config is passed to NewCache for creating new Cache instances.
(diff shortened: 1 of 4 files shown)


Darkweak:
Up @martinmr

anurags92:
@Darkweak @martinmr is on leave, let me bring this up internally and get back.

Darkweak:
@jarifibrahim @manishrjain can you review please ?

cache.go
```diff
@@ -255,6 +275,10 @@ func (c *Cache) SetWithTTL(key, value interface{}, cost int64, ttl time.Duration
 		return false
 	}
 
+	if c.keySaver != nil {
+		c.keySaver.AddKey(fmt.Sprintf("%v", key))
```
jarifibrahim, line 279: It could be possible that the item is rejected by the cache if after this point if
1. We're under heavy load or
2. The cost of item exceeds the max cost of the cache.

This isn't the right place to Add the key to the keySaver.
@Darkweak I think the keysaver stuff doesn't need to be in ristretto. You can keep the keysaver in your application code.
Here's how you would do it
```
cache.Evict = func(item *ristretto.Item) { keySaver.Del(item.Key()) }
if cache.Set(...,...) { keySaver.Add(....) }
```

jarifibrahim (changes requested):
Thanks for the PR @Darkweak! Please see my comment.

cache.go
Darkweak: Thank you for your response, IMO we can close this PR and i'll implement your solution and support it directly in my application. You're right, ristretto shouldn't support that and could decrease performances too.

Thanks for the work you're doing on ristretto 👍

jiamo: Why item only have keyHash. The key was lost. Annd when `cache.Set` we can't get keyHash from cache. So how can I got key by `item.Key()`?

PR #258, opened 2021-03-09
feat(super-flags): Add GetPath method in superflags

ahsanbarkati:
In super-flags, the file-paths are given within quotes. So the shell is not able to expand the paths having `~` to `/home/user/...`.
This PR adds `GetPath()` method on super flags which expands the paths having `~` to `/home/user/..`.

For example:
`dgraph alpha --acl="secret_file=~/hmac-secret"` 

the above command throws an error that path doesn't exists even if it exists.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/258)

### DIFF
--- a/z/flags.go
+++ b/z/flags.go
@@ -259,13 +259,13 @@ func (sf *SuperFlag) GetPath(opt string) string {
 	return path
 }
 
-// expandPath expands the paths containing ~ to /home/user. It also computes the ablosule path
+// expandPath expands the paths containing ~ to /home/user. It also computes the absolute path
 // from the relative paths. For example: ~/abc/../cef will be transformed to /home/user/cef.
 func expandPath(path string) (string, error) {
 	if len(path) == 0 {
 		return "", nil
 	}
-	if path[0] == '~' {
+	if path[0] == '~' && (len(path) == 1 || path[1] == '/') {
 		usr, err := user.Current()
 		if err != nil {
 			return "", errors.Wrap(err, "Failed to get the home directory of the user")
--- a/z/flags.go
+++ b/z/flags.go
@@ -3,6 +3,8 @@ package z
 import (
 	"fmt"
 	"log"
+	"os/user"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


karlmcguire (approved):
:lgtm:

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/258#-:-MVMXO2t5DdmU-UlZMMt:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jarifibrahim, @manishrjain, and @martinmr)

karlmcguire:
Does this address row #34 in Haqathon sheet?

ahsanbarkati:
> Does this address row #34 in Haqathon sheet? If not, might be a good place to improve that.

If we want to support filenames with semi-colons then we will have to change the separator itself. Not sure if we expect filenames with semi-colons.

z/flags.go
```diff
@@ -247,3 +249,34 @@ func (sf *SuperFlag) GetString(opt string) string {
...
+func (sf *SuperFlag) GetPath(opt string) string {
+	p := sf.GetString(opt)
+	path, err := expandPath(p)
+	if err != nil {
+		log.Fatalf("Failed to get path: %+v", err)
+	}
+	return path
+}
+
+// expandPath expands the paths containing ~ to /home/user. It also computes the ablosule path
```
jarifibrahim, line 262: typo ablosule => absolute

jarifibrahim (approved):

z/flags.go
```diff
@@ -247,3 +249,34 @@ func (sf *SuperFlag) GetString(opt string) string {
...
+// from the relative paths. For example: ~/abc/../cef will be transformed to /home/user/cef.
+func expandPath(path string) (string, error) {
+	if len(path) == 0 {
+		return "", nil
+	}
+	if path[0] == '~' {
```
ajeetdsouza, line 268: This does not take into account filenames that start with `~`, eg. `"~myfilename"`.

On UNIX,
```go
path[0] == '~' && (len(path) == 1 || path[1] == '/')
```

On Windows,
```go
path[0] == '~' && (len(path) == 1 || path[1] == '/' || path[1] == '\\')
```

ajeetdsouza (changes requested):

z/flags.go
ahsanbarkati: Done

z/flags.go
ahsanbarkati: We can actually create file with names `'~\\filename'` in unix. Which will cause problems if we include windows fix. I have fixed this for unix, but not for windows.

ajeetdsouza: @ahsanbarkati you can use `os.IsPathSeparator` to do this in a cross-platform manner.

ahsanbarkati: Nice. Thanks. Done :)

PR #268, opened 2021-04-21
Use z.Buffer backing for B+ tree

ajeetdsouza:
- `z.Tree` should use a `z.Buffer` backing.
- `z.Buffer` should use a `z.MmapFile` backing.
- Add persistent API to `z.Buffer`, so it doesn't delete the file after closing.
- Add persistent API to `z.Tree`, so that we can use it for persistent storage.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/268)

### DIFF
--- a/z/btree.go
+++ b/z/btree.go
@@ -59,6 +59,18 @@ func NewTree() *Tree {
 	return t
 }
 
+// NewTree returns a persistent on-disk B+ tree.
+func NewTreePersistent(path string) (*Tree, error) {
+	file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0666)
+	if err != nil {
+		return nil, err
+	}
+	buffer := NewBufferFromFile(file, 0).WithPersistent(true)
+	tree := &Tree{buffer: buffer}
+	tree.Reset()
+	return tree, nil
+}
+
 // Reset resets the tree and truncates it to maxSz.
 func (t *Tree) Reset() {
 	t.nextPage = 1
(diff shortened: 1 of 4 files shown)


z/btree.go
```diff
@@ -51,22 +52,44 @@ func (t *Tree) initRootNode() {
 	t.Set(absoluteMax, 0)
 }
 
-// NewTree returns a memory mapped B+ tree with given filename.
+// NewTree returns an in-memory B+ tree.
 func NewTree() *Tree {
-	t := &Tree{}
+	t := &Tree{buffer: NewBuffer(minSize, "tree")}
```
jarifibrahim, line 57: The NewBuffer by default uses calloc. The current implementation was using go memory. With this change, we will use calloc'ed memory instead of go memory for the btree.

z/buffer.go
```diff
@@ -183,56 +143,72 @@ func (b *Buffer) Bytes() []byte {
...
 // expansion.
 func (b *Buffer) Grow(n int) {
-	// In this case, len and cap are the same.
 	if b.buf == nil {
 		panic("z.Buffer needs to be initialized before using")
 	}
-	if b.maxSz-int(b.offset) < n {
-		panic(fmt.Sprintf("Buffer max size exceeded: %d."+
-			" Offset: %d. Grow: %d", b.maxSz, b.offset, n))
+	if b.maxSz != 0 && int(b.offset)+n > b.maxSz {
```
jarifibrahim, line 154: When can b.maxSz be 0? It should always have a valid value.

ajeetdsouza: I've kept the zero value as an "unbounded" buffer. The earlier API set `maxSz = math.MaxInt32` (roughly 2.1GB) by default. Should I change it back?

manishrjain (approved):
:lgtm:

> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/268#-:-MYuvB_QENCOY9xrQ7_s:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @ajeetdsouza and @jarifibrahim)

z/btree.go
ajeetdsouza: Asked Manish, he said Calloc should be fine.

PR #271, opened 2021-04-29
Initialize persistent tree correctly

ajeetdsouza:
`PersistentTree` currently does not set `t.freePage` and `t.nextPage` correctly upon opening an existing file, leading to errors. Additionally, stats are also lost. This PR computes these values when the tree is initialized.

TODO:
- [x] Add tests for `PersistentTree`

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/271)

### DIFF
--- a/z/btree.go
+++ b/z/btree.go
@@ -28,9 +28,12 @@ import (
 )
 
 var (
-	pageSize    = os.Getpagesize()
-	maxKeys     = (pageSize / 16) - 1
-	oneThird    = int(float64(maxKeys) / 3)
+	pageSize = os.Getpagesize()
+	maxKeys  = (pageSize / 16) - 1
+	oneThird = int(float64(maxKeys) / 3)
+)
+
+const (
 	absoluteMax = uint64(math.MaxUint64 - 1)
 	minSize     = 1 << 20
 )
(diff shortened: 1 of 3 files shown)


z/btree.go
```diff
@@ -65,24 +68,84 @@ func NewTree(tag string) *Tree {
...
 	}
-	tree := &Tree{buffer: buffer}
-	tree.Reset()
-	return tree, nil
+	t.buffer.offset = uint64(len(t.buffer.buf))
+	t.data = t.buffer.Bytes()
+
+	// TODO(ajeet): can we do better than this?
+	isInitialized := false
+	for _, b := range t.data {
```
NamanJain8, line 84: You can just look at the root node, t.node(1), and check if that is initialized or not. If any part like pageID, numKeys, etc is non-zero for that node, then it is initialized.

z/btree.go
```diff
@@ -65,24 +68,84 @@ func NewTree(tag string) *Tree {
...
+		if b != 0 {
+			isInitialized = true
+			break
+		}
+	}
+	if !isInitialized {
+		t.nextPage = 1
+		t.freePage = 0
+		return t, nil
```
NamanJain8, line 93: You also need to call `t.initRootNode()`.

z/btree.go
```diff
@@ -65,24 +68,84 @@ func NewTree(tag string) *Tree {
...
+		}
+	}
+	// Mark all pages being pointed to as used.
+	for _, pageId := range pointedPages {
+		i := pageId - 1
+		usedPages[i] = true
+	}
+	// Only one page would still be marked as free.
+	// This is the root free page node.
+	for i, used := range usedPages {
```
NamanJain8, line 132: `usedPages` is slightly confusing name. Can we name it better?

NamanJain8 (changes requested):

z/btree.go
ajeetdsouza: Since `pageID` can never be zero, we can just check for that, right? We shouldn't need to check anything else.

z/btree.go
ajeetdsouza: Hm, maybe. What should I name it to? I could name it `freePages` and flip all the bits, but that seems unnecessary.

z/btree.go
NamanJain8: Correct. This check is enough.

z/btree.go
```diff
@@ -65,24 +68,86 @@ func NewTree(tag string) *Tree {
...
+
+	if !isInitialized {
+		t.nextPage = 1
+		t.freePage = 0
+		t.initRootNode()
+		return t, nil
+	}
+
+	// Calculate t.nextPage by finding the highest pageId among all the nodes.
+	t.nextPage = 0
```
jarifibrahim, line 94: Move this into a new function.
```
if newTree {
 ... 
return
}

t.reinit()
```

z/btree_test.go
```diff
@@ -59,6 +60,35 @@ func TestTree(t *testing.T) {
...
+	dir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	path := filepath.Join(dir, "tree.buf")
+
+	// Create a tree and validate the data.
+	bt, err := NewTreePersistent(path)
+	require.NoError(t, err)
+	N := uint64(256 * 256)
```
jarifibrahim, line 73: 64<<10

z/btree_test.go
```diff
@@ -59,6 +60,35 @@ func TestTree(t *testing.T) {
...
+	for i := uint64(1); i < N; i++ {
+		bt.Set(i, i*2)
+	}
+	for i := uint64(1); i < N; i++ {
+		require.Equal(t, i*2, bt.Get(i))
+	}
+	require.NoError(t, bt.Close())
```
jarifibrahim, line 80: Check all the stats as well.

jarifibrahim (approved):
Defer to @NamanJain8 for final approval.

z/btree.go
ajeetdsouza: Renamed to `childPages`.

PR #287, opened 2021-10-19
fix runtime error: index out of range for !amd64 env

awsl-dbq:
fix `runtime error: index out of range`  in non-amd64 (arm64) env

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/287)


awsl-dbq:
if Len(xs) mod 8 !=0 , then xs[i+2] may got index out of range error

PR #295, opened 2022-05-26
make policy an interface

tushar-zomato:
add option to always admit new items

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -14,4 +14,4 @@ enabled = true
 
 
   [analyzers.meta]
-  import_path = 'github.com/dgraph-io/ristretto'
+  import_path = 'github.com/tushar-zomato/ristretto'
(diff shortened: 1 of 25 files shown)


joshua-goldstein:
There is the option of setting `MaxCost` and `NumCounters` to be very large values, so that the cache will admit essentially all items (except those dropped in the buffers).  You might also be able to add items with the cost of each item to be 0.  These are much easier solutions than adding an entirely new policy.

joshua-goldstein:
There is the option of setting `MaxCost` and `NumCounters` to be very large values, so that the cache will admit essentially all items (except those dropped in the buffers).  You might also be able to add items with the cost of each item to be 0.  These are much easier solutions than adding an entirely new policy.

.deepsource.toml
```diff
@@ -14,4 +14,4 @@ enabled = true
 
 
   [analyzers.meta]
-  import_path = 'github.com/dgraph-io/ristretto'
+  import_path = 'github.com/tushar-zomato/ristretto'
```
MichelDiz, line 17: Why this renaming?

MichelDiz (changes requested):

PR #298, opened 2022-08-18
add test to ensure calling set after close does not panic

joshua-goldstein:
In a prior version of Ristretto, calling set after close could cause a panic.  We are adding this test to check for this behavior and prevent a regression.  Test was originally written by @josephschorr in [this PR](https://github.com/outcaste-io/ristretto/pull/5).

### DIFF
--- a/cache_test.go
+++ b/cache_test.go
@@ -1004,3 +1004,51 @@ func TestCacheWithTTL(t *testing.T) {
 		})
 	}
 }
+
+func TestConcurrentSetAfterClose(t *testing.T) {
+	c, err := NewCache(&Config{
+		NumCounters:        100,
+		MaxCost:            10,
+		IgnoreInternalCost: true,
+		BufferItems:        64,
+		Metrics:            true,
+	})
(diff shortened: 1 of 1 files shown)


cache_test.go
```diff
@@ -997,3 +997,51 @@ func TestCacheWithTTL(t *testing.T) {
...
+	tc := time.Tick(50 * time.Millisecond)
+	<-tc
+
+	// Run close to ensure nothing panics.
+	c.Close()
+
+	// Wait some more time so Set can run a few additional times.
+	tcd := time.Tick(250 * time.Millisecond)
+	<-tcd
+	close(done)
```
matthewmcneely, line 1046: @joshua-goldstein Is there some state that can be checked to see if the test succeeded? Other than the check of `err` on the creation of `NewCache`, there's no other way the test can fail, right?

josephschorr: The idea behind the test is that if the bug exists, the race detector will raise a panic (thus breaking the test)

PR #299, opened 2022-08-18
Joshua/fix .go version

joshua-goldstein:
test.sh has a variable GOVERSION that checks if there is a file .go-version and sets GOVERSION to the value specified in the file.  This is then used as a tag for the appropriate Docker image of Go.  The previous value was go1.17.11 which is not a valid tag in the [Go Docker hub](https://hub.docker.com/_/golang/tags?page=1&name=1.17.11) (it should omit the prefix go).  This was causing the test to fail.


.go-version
```diff
@@ -1 +1 @@
-go1.17.11
+1.17.11
```
skrdgraph, line 1: nice catch @joshua-goldstein ... now the tests are running (and progressing further than before)

skrdgraph (approved):
👍

PR #305, opened 2022-10-09
Reset all arrays at once

falsandtru:
Decrease the latency on reset.

## Problem

The current implementation iterates arrays for each array.

## Solution

Iterate all arrays at once.

### DIFF
--- a/sketch.go
+++ b/sketch.go
@@ -82,15 +82,21 @@ func (s *cmSketch) Estimate(hashed uint64) int64 {
 
 // Reset halves all counter values.
 func (s *cmSketch) Reset() {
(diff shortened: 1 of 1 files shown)


falsandtru:
CLAassistant requires excessive permissions.

>Gists
Read and write access
Organizations and teams
Read-only access
Commit statuses
Read and write access
Repository webhooks and services
Read and write access

joshua-goldstein:
Hi @falsandtru , is there a benchmark that shows the performance gain when iterating in this way?

falsandtru:
I don't know how to benchmark the performance.

PR #309, opened 2022-10-10
fix(z): Address inconsistent mremap return arguments with arm64

matthewmcneely:
This PR adds a specific arm64 implementation for the `mremap` function which ignores the second return argument from the syscall. **All other Linux implementations are unchanged.**

This PR reworks [an earlier PR](https://github.com/dgraph-io/ristretto/pull/281) so that the functionality on non-arm64 platforms remains the same.

Tested on Linux amd64 and Linux arm64.

### DIFF
--- a/z/mremap_linux_arm64.go
+++ b/z/mremap_linux_arm64.go
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2020 Dgraph Labs, Inc. and Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


z/mremap_linux_arm64.go
```diff
@@ -0,0 +1,52 @@
+/*
```
joshua-goldstein, line 1: Should we have a build directive like `//  +build arm64` here?

matthewmcneely: Not needed, the `_arm64` suffix is recognized by the go compiler

joshua-goldstein: I see, good to know

joshua-goldstein (approved):
👍

PR #319, opened 2022-12-06
chore(ci): pin runner to ubuntu 20.04

joshua-goldstein:
## Problem

Lint tests and CI tests were failing.

## Solution

Pin ristretto test runner to Ubuntu 20.04, and bump up linter dependencies.


adityasadalage (approved):
Looks good

PR #321, opened 2022-12-11
feat(Generics): add generics.

aryehlev:
## Problem
not using generics.
 -->

## Solution
add generics.
 -->

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -28,7 +28,7 @@ import (
 	"time"
 	"unsafe"
 
-	"github.com/aryehlev/ristretto/z"
+	"github.com/dgraph-io/ristretto/z"
 )
 
 var (
@@ -50,7 +50,7 @@ func Zero[T any]() T {
 // Cache is a thread-safe implementation of a hashmap with a TinyLFU admission
 // policy and a Sampled LFU eviction policy. You can use the same Cache instance
 // from as many goroutines as you want.
-type Cache[K comparable, V any] struct {
+type Cache[K any, V any] struct {
 	// store is the central concurrent hashmap where key-value items are stored.
 	store store[V]
 	// policy determines what gets let in to the cache and what gets kicked out.
@@ -70,7 +70,7 @@ type Cache[K comparable, V any] struct {
 	// KeyToHash function is used to customize the key hashing algorithm.
 	// Each key will be hashed using the provided function. If keyToHash value
 	// is not set, the default keyToHash function is used.
-	keyToHash func(any) (uint64, uint64)
+	keyToHash func(K) (uint64, uint64)
 	// stop is used to stop the processItems goroutine.
 	stop chan struct{}
 	// indicates whether cache is closed.
@@ -88,7 +88,7 @@ type Cache[K comparable, V any] struct {
 }
 
 // Config is passed to NewCache for creating new Cache instances.
-type Config[K comparable, V any] struct {
+type Config[K any, V any] struct {
 	// NumCounters determines the number of counters (keys) to keep that hold
 	// access frequency information. It's generally a good idea to have more
 	// counters than the max cache capacity, as this will improve eviction
@@ -164,7 +164,7 @@ type Item[V any] struct {
 }
 
 // NewCache returns a new Cache instance and any configuration errors, if any.
-func NewCache[K comparable, V any](config *Config[K, V]) (*Cache[K, V], error) {
+func NewCache[K any, V any](config *Config[K, V]) (*Cache[K, V], error) {
 	switch {
 	case config.NumCounters == 0:
 		return nil, errors.New("NumCounters can't be zero")
@@ -175,11 +175,11 @@ func NewCache[K comparable, V any](config *Config[K, V]) (*Cache[K, V], error) {
 	}
 	policy := newPolicy[V](config.NumCounters, config.MaxCost)
 	cache := &Cache[K, V]{
-		store:  newStore[V](),
-		policy: policy,
-		getBuf: newRingBuffer(policy, config.BufferItems),
-		setBuf: make(chan *Item[V], setBufSize),
-		//keyToHash:          config.KeyToHash,
+		store:              newStore[V](),
+		policy:             policy,
+		getBuf:             newRingBuffer(policy, config.BufferItems),
+		setBuf:             make(chan *Item[V], setBufSize),
+		keyToHash:          config.KeyToHash,
 		stop:               make(chan struct{}),
 		cost:               config.Cost,
 		ignoreInternalCost: config.IgnoreInternalCost,
@@ -203,7 +203,7 @@ func NewCache[K comparable, V any](config *Config[K, V]) (*Cache[K, V], error) {
 		cache.onExit(item.Value)
 	}
 	if cache.keyToHash == nil {
-		cache.keyToHash = z.KeyToHash
+		cache.keyToHash = z.KeyToHash[K]
 	}
 	if config.Metrics {
 		cache.collectMetrics()
@@ -212,6 +212,7 @@ func NewCache[K comparable, V any](config *Config[K, V]) (*Cache[K, V], error) {
 	//       goroutines we have running cache.processItems(), so 1 should
 	//       usually be sufficient
 	go cache.processItems()
+
 	return cache, nil
 }
 
@@ -230,7 +231,7 @@ func (c *Cache[K, V]) Wait() {
 // the same time.
 func (c *Cache[K, V]) Get(key K) (V, bool) {
 
-	if c == nil || c.isClosed || key == Zero[K]() {
+	if c == nil || c.isClosed || &key == nil {
 		return Zero[V](), false
 	}
 	keyHash, conflictHash := c.keyToHash(key)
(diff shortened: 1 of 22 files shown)


GRbit:
@aryehlev since there is no reaction from the repository maintainer, have you thought about updating your fork with a description with a headline like:
```
This is a fork of github.com/dgraph-io/ristretto but with **Magick of Generics**
```
I think it's an incredibly convenient feature to have a ristretto with generics, so it has the potential of getting some stars.

mangalaman93:
Thanks for getting this to our attention. I will review the PR now. Please sign the CLA when you can.

cache.go
```diff
@@ -36,40 +36,47 @@ var (
 	setBufSize = 32 * 1024
 )
 
-type itemCallback func(*Item)
+const itemSize = int64(unsafe.Sizeof(storeItem[any]{}))
 
-const itemSize = int64(unsafe.Sizeof(storeItem{}))
+//type HashableKey interface {
```
mangalaman93, line 41: please remove this

cache.go
```diff
@@ -36,40 +36,47 @@ var (
...
+//	KeyToHash() (uint64, uint64)
+//}
+
+func Zero[T any]() T {
```
mangalaman93, line 45: can we not export this function and rename it to something else? Zero is a well defined node in Dgraph. Maybe call it zeroValue?

cache.go
```diff
@@ -36,40 +36,47 @@ var (
...
 // policy and a Sampled LFU eviction policy. You can use the same Cache instance
 // from as many goroutines as you want.
-type Cache struct {
+type Cache[K any, V any] struct {
 	// store is the central concurrent hashmap where key-value items are stored.
-	store store
+	store store[V]
 	// policy determines what gets let in to the cache and what gets kicked out.
-	policy policy
+	policy policy[V]
```
mangalaman93, line 57: same, field name is same as type name

cache.go
```diff
@@ -205,25 +212,27 @@ func NewCache(config *Config) (*Cache, error) {
...
 }
 
 // Get returns the value (if any) and a boolean representing whether the
 // value was found or not. The value can be nil and the boolean can be true at
 // the same time.
-func (c *Cache) Get(key interface{}) (interface{}, bool) {
-	if c == nil || c.isClosed || key == nil {
-		return nil, false
+func (c *Cache[K, V]) Get(key K) (V, bool) {
+
```
mangalaman93, line 233: nit: remove the extra new line

cache.go
```diff
@@ -205,25 +212,27 @@ func NewCache(config *Config) (*Cache, error) {
...
+	if c == nil || c.isClosed || &key == nil {
```
mangalaman93, line 234: pointer to key can never be nil. That is why I am wondering whether we should take a pointer to K (*K) as the arg to Get function

cache.go
```diff
@@ -205,25 +212,27 @@ func NewCache(config *Config) (*Cache, error) {
...
+		return Zero[V](), false
```
mangalaman93, line 235: same, return value could also be of type *V

mangalaman93 (changes requested):
I think we need to figure out if we want to use pointer to K & V everywhere first. I can review the rest of the PR then. Let me know what you think. Thanks for the PR.

cache.go
```diff
@@ -166,37 +173,37 @@ func NewCache(config *Config) (*Cache, error) {
...
+		setBuf:             make(chan *Item[V], setBufSize),
 		keyToHash:          config.KeyToHash,
 		stop:               make(chan struct{}),
 		cost:               config.Cost,
 		ignoreInternalCost: config.IgnoreInternalCost,
 		cleanupTicker:      time.NewTicker(time.Duration(bucketDurationSecs) * time.Second / 2),
 	}
-	cache.onExit = func(val interface{}) {
-		if config.OnExit != nil && val != nil {
+	cache.onExit = func(val V) {
```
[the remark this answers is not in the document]
aryehlev: @mangalaman93 @GRbit i think my initial design I used pointers in a lot of these places and ended up feeling it was redundant, I will look at it again later

aryehlev:
@GRbit @mangalaman93 

Hey, thank you for the comments.
I can definitely look at this and do the necessary fixes later. I think I did not use pointers for a reason.

It is definitely a major change and will only work with go 1.18 and up so I am not sure how you will want to release this.

mangalaman93:
We have upgraded badger and dgraph to Go 1.19, we should be able to do that for ristretto too while we do the next major release. It should be fine to test the PR with the same.

cache.go
```diff
@@ -205,29 +208,35 @@ func NewCache(config *Config) (*Cache, error) {
...
+func (c *Cache[K, V]) Get(key K) (V, bool) {
+	if c == nil || c.isClosed {
+		return zeroValue[V](), false
 	}
 	keyHash, conflictHash := c.keyToHash(key)
+
+	// If key in a pointer type and nil the default keyToHash function will return 0,0
+	if keyHash == 0 && conflictHash == 0 {
+		return zeroValue[V](), false
+	}
```
aryehlev, line 237: i agree that &key == nil is redundant but a pointer to K (*K) as the arg is not the right way to go imo.

@mangalaman93 if key is nil then 
`keyHash, conflictHash := c.keyToHash(key)`
i have added that will return 0,0 in the default keyToHash, if users want to implement their own keyToHash and also use a pointer type as their key they should implement the function to return 0,0 in case of nil. WDYT?

cache.go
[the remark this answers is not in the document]
mangalaman93: @aryehlev did you have a chance to look at it again? This PR looks really good to me and we can merge it soon.

mangalaman93:
Let me know if you are facing any challenge resolving conflicts.

aryehlev:
i think i now resolved all conflicts,  please go over the code if u can

store.go
```diff
@@ -35,58 +35,58 @@ type storeItem struct {
...
 	// Set adds the key-value pair to the Map or updates the value if it's
 	// already present. The key-value pair is passed as a pointer to an
 	// item object.
-	Set(*Item)
+	Set(*Item[V])
 	// Del deletes the key-value pair from the Map.
-	Del(uint64, uint64) (uint64, interface{})
-	// Update attempts to update the key with a new value and returns true if
+	Del(uint64, uint64) (uint64, V)
+	// Update attempxsts to update the key with a new value and returns true if
```
mangalaman93, line 49: attempts

z/z.go
```diff
@@ -18,22 +18,46 @@ package z
...
-// second hash (see Ristretto PR #88).
-// We also know that while the Go runtime has a runtime memhash128
-// function, it's not possible to use it to generate [2]uint64 or
-// anything resembling a 128bit hash, even though that's exactly what
-// we need in this situation.
-func KeyToHash(key interface{}) (uint64, uint64) {
-	if key == nil {
-		return 0, 0
+// GetKeyToHash will get the default KeyToHash function for the primitive types.
+func GetKeyToHash[K any](emptyKey K) func(K) (uint64, uint64) {
```
mangalaman93, line 28: I am wondering if it is possible to define this function at compile time somehow based on the type parameter. Is there a way we could achieve that using Go generics?

mangalaman93 (changes requested):
A few more comments. I will ask more engineers to review this PR.

z/z.go
aryehlev: i couldnt think of an elegant way since the KeyToHash function is expected to also be generic,
but let me think a little more

aryehlev: @mangalaman93 there is no way i can find make this function generic sadley:(

z/z.go
```diff
@@ -18,22 +18,47 @@ package z
...
+	default:
+		return nil, errors.New("must provide custom KeyToHash function for given type")
 	}
+
+	return KeyToHash[K], nil
+}
+
+// TODO: Figure out a way to re-use memhash for the second uint64 hash, we
+//
+//	already know that appending bytes isn't reliable for generating a
```
mangalaman93, line 53: you have to fix the double space after the // so that the whole TODO is together.
Additionally, I am thinking we don't need to change anything in the file. The original function works as is?

policy.go
```diff
@@ -33,13 +33,14 @@ const (
 // policy is the interface encapsulating eviction/admission behavior.
 //
 // TODO: remove this interface and just rename defaultPolicy to policy, as we
-// are probably only going to use/implement/maintain one policy.
-type policy interface {
+//
+//	are probably only going to use/implement/maintain one policy.
```
mangalaman93, line 37: the same here as well. This was done by `go fmt` but if you remove the space it won't add the additional line.

mangalaman93 (changes requested):
Hi @aryehlev, just a couple of minor comments and we are good to merge. Thanks for all the work.

z/z.go
aryehlev: so two things about the original function.
1: we must wrap it with this funciton. 
`func KeyToHash[K any](key K) (uint64, uint64) {
	return defaultKeyToHash(key)
}`
2: i want to catch also on initialization if there is no function given for other types and thats why i want this function added.
`func GetKeyToHash[K any](emptyKey K) (func(K) (uint64, uint64), error)`
but i can theoretically remove this.

store.go
mangalaman93: I thought this change was resolved

mangalaman93 (approved):
Looks good to me now! Thank you so much for all the work. I see one typo/spell error, I thought you had fixed it. Could you check that again please? I also want to get more reviews on this, let me see who is available.

GRbit:
I would like to congratulate @aryehlev on getting this PR approved and thank you both for your work! This was a long PR and it was nice to see it get this far!

PR #325, opened 2022-12-23
fix(cmSketch test) and add better cmSketch test

FrankReh:
One Sketch test was intended to check that the seeds had caused each sketch row to be unique but the way the test was iterating, the failure wasn't going to be triggered.

    With this fix, the test passes seeming to indicate the seeds are
    working as intended on the row creation. But there is a problem.

A new Sketch test is added that goes to the trouble of sorting the counters of each row to ensure that each row isn't just essentially a copy of another row, where the only difference is which position the counters occupy.

And the new Sketch test is performed twice, once with high-entropy hashes, because they come from a PRNG, and once with low-entropy hashes, which in many cases will be normal, because they are all small numbers, good for indexing, but not good for getting a spread when simply applied to a bitwise XOR operation.

    These new tests show a problem with the counter increment logic
    within the cmSketch.Increment method which was most likely
    introduced by commit f823dc4a.

    A subsequent commit addresses the problems surfaced. But as the
    discussion from issue #108 shows (discussion later moved to
    https://discuss.dgraph.io/t/cmsketch-not-benefitting-from-four-rows/8712 ),
    other ideas for incrementing the counters were considered
    by previous authors as well.

Fix existing cmSketch test and add improved cmSketch test

(Marked as draft because this introduces a test that fails for now. I can commit a fix to the cmSketch increment method to get the new test to pass - if a maintainer agrees there is a problem to be fixed. See #108. I tried a few years ago.)

One Sketch test was intended to check that the seeds had caused each sketch row to be unique but the way the test was iterating, the failure wasn't going to be triggered.

    With this fix, the test passes seeming to indicate the seeds are
    working as intended on the row creation. But there is a problem with
    the actual increment method. A new Sketch test is added that goes
    further and sorts the counters of each row to ensure that each row
    isn't just essentially a copy of another row, where the only
    difference is which position the counters occupy.

And the new Sketch test is performed twice, once with high-entropy hashes, because they come from a PRNG, and once with low-entropy hashes, which in many cases will be normal, because they are all small numbers, good for indexing, but not good for getting a spread when simply applied with a bitwise XOR operation.

    These new tests show a problem with the counter increment logic
    within the cmSketch.Increment method which was most likely
    introduced by commit f823dc4a.

    A subsequent commit addresses the problems surfaced. But as the
    discussion from issue #108 shows (discussion later moved to
    https://discuss.dgraph.io/t/cmsketch-not-benefitting-from-four-rows/8712),
    other ideas for incrementing the counters were considered by
    previous authors of the original Java code as well.

### DIFF
--- a/sketch_test.go
+++ b/sketch_test.go
@@ -1,6 +1,9 @@
 package ristretto
 
 import (
+	"math/rand"
+	"sort"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


FrankReh:
It's too bad things like this aren't being fixed. I've seen the logic error copied as part of other language ports.

FrankReh:
Anybody?

FrankReh:
I hope it is clear to anyone reviewing that this test is failing because of the logic bug I pointed out years ago. I had introduced a separate PR to fix that. This PR just highlights the problem more clearly for someone new.

FrankReh:
Would like to keep this open until a maintainer can address it or say they don't care. Thanks!

mangalaman93:
@FrankReh Thank you for your PR. I understand this maybe frustrating but we appreciate the change and will look at it as soon as we can. This is under our watch and we need to spend some time on it for reviewing it. Let me see if we can prioritize it sooner.

FrankReh:
Just wanted to keep this alive so others can still see it if they look at this crate.

FrankReh:
Anyone?

PR #342, opened 2023-07-18
`feat(configurable cleanup ticker)`: add config for cleanup ticker duration

singhvikash11:
## Problem
Cleanup ticker to remove entries for expired keys taking a lot of CPU cycles around 90% in the case of pod with 1.6 CPU

## Solution
 Making the duration of the cleanup ticker configurable so users can configure it to higher values and can save on CPU cycles.


singhvikash11:
@joshua-goldstein can you please review the pr, I want to use this library as app-cache in prod.

singhvikash11:
@akon-dey can you please review the pr

mangalaman93 (approved):

PR #355, opened 2023-09-03
feat: load value if not present

proost:
Resolve: https://github.com/dgraph-io/ristretto/issues/332

Get value if cache hit, or load value, set to cache and return loaded value.

only one go-routine can execute load function per key.


proost:
@joshua-goldstein @mangalaman93 @harshil-goel @all-seeing-code
can you give me any feedback? If this feature is unnecessary, I will close the PR.

PR #281, opened 2021-08-20
Remove mmapSize check, it has wrong value on ARM64

yarysh:
In this statement:
`mmapAddr, mmapSize, errno := unix.Syscall6(unix.SYS_MREMAP,...)`

`mmapSize` - it is not an allocated size, at least for ARM64 arch.
For ARM64, `mmapSize` will be equal to the old size.

Example script: https://gist.github.com/yarysh/2b4db19ebe1236f31ff6780d956628f5

I run this script on **AMD64** machine with Ubuntu 20.04, and it returns:
```
Requested size: 20
mmapAddr: 824634507264
mmapSize: 20
```

And on **ARM64** machine (Raspberry PI and AWS Graviton) with Ubuntu 20.04, it returns:
```
Requested size: 20
mmapAddr: 274878390272
mmapSize: 1024
mremap size mismatch: requested: 20 got: 1024
```

I have checked man for mremap and it describes returns as:
```
RETURN VALUE
       On  success mremap() returns a pointer to the new virtual memory area.  On error,
       the value MAP_FAILED (that is, (void *) -1) is returned, and errno is set appropriately.
```

This issue also affected **badger** on ARM64: https://discuss.dgraph.io/t/error-mremap-size-mismatch-on-arm64/15333

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/ristretto/281)

PR #366, opened 2023-12-09
Disable mmap size check on arm arch as well as arm64

requilence:
In #281 mmap size check has been disabled for the ARM64 arch. 

But we at [anytype](https://github.com/anyproto) quite often face the same problem on the arm arch on Android devices (armeabi-v7a target) with the [latest badger v4.2.0](https://github.com/dgraph-io/badger/releases/tag/v4.2.0) and [latest ristretto v0.1.1](https://github.com/dgraph-io/ristretto).

```memtables error: while opening fid: 1 error: while updating skiplist error: mremap size mismatch: requested: 1479692 got: 67108864```

This PR extends the #281 to cover all arm architectures.

Here is one more mention of this issue
https://github.com/dgraph-io/badger/issues/2027#issuecomment-1820368978

It would also make sense to create the PR on top of v0.1.1 release, but there is not release branch to base on

### DIFF
--- a/z/mremap_nosize.go
+++ b/z/mremap_nosize.go
@@ -1,5 +1,6 @@
-//go:build !arm64
-// +build !arm64
+//go:build (arm64 || arm) && linux
+// +build arm64 arm
+// +build linux
 
 /*
  * Copyright 2020 Dgraph Labs, Inc. and Contributors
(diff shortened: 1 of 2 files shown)


mangalaman93 (approved):
LGTM

PR #371, opened 2024-01-16
Restrict generic key type to only those supported

pete-woods:
Without this, you can specify key types that aren't supported and then have to deal with panics, rather than more pleasant compile-time warnings.

### DIFF
--- a/z/z.go
+++ b/z/z.go
@@ -23,6 +23,10 @@ import (
 	"github.com/cespare/xxhash/v2"
 )
 
+type Key interface {
+	uint64 | string | []byte | byte | int | int32 | uint32 | int64
+}
+
 // TODO: Figure out a way to re-use memhash for the second uint64 hash,
 // we already know that appending bytes isn't reliable for generating a
 // second hash (see Ristretto PR #88).
@@ -30,11 +34,8 @@ import (
 // function, it's not possible to use it to generate [2]uint64 or
 // anything resembling a 128bit hash, even though that's exactly what
 // we need in this situation.
-func KeyToHash[K any](key K) (uint64, uint64) {
+func KeyToHash[K Key](key K) (uint64, uint64) {
 	keyAsAny := any(key)
-	if keyAsAny == nil {
-		return 0, 0
-	}
 	switch k := keyAsAny.(type) {
 	case uint64:
 		return k, 0
--- a/cache.go
+++ b/cache.go
@@ -43,10 +43,12 @@ func zeroValue[T any]() T {
 	return zero
 }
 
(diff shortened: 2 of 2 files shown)


mangalaman93 (approved):

mangalaman93:
Thanks for the PR. The only problem that I see with this is that any new type addition requires changes to the code and upgrading ristretto.

pete-woods:
> Thanks for the PR. The only problem that I see with this is that any new type addition requires changes to the code and upgrading ristretto.

I believe this is already the case (see linked code): https://github.com/dgraph-io/ristretto/blob/main/z/z.go#L39

z/z.go
```diff
@@ -23,18 +23,19 @@ import (
...
 // we already know that appending bytes isn't reliable for generating a
 // second hash (see Ristretto PR #88).
 // We also know that while the Go runtime has a runtime memhash128
 // function, it's not possible to use it to generate [2]uint64 or
 // anything resembling a 128bit hash, even though that's exactly what
 // we need in this situation.
-func KeyToHash[K any](key K) (uint64, uint64) {
+func KeyToHash[K Key](key K) (uint64, uint64) {
 	keyAsAny := any(key)
-	if keyAsAny == nil {
```
pete-woods, line 35: I *think* the linter correctly identified this check as redundant now (it forced me to remove it).

z/z.go
```diff
@@ -23,18 +23,19 @@ import (
...
 // TODO: Figure out a way to re-use memhash for the second uint64 hash,
 // we already know that appending bytes isn't reliable for generating a
 // second hash (see Ristretto PR #88).
 // We also know that while the Go runtime has a runtime memhash128
 // function, it's not possible to use it to generate [2]uint64 or
 // anything resembling a 128bit hash, even though that's exactly what
 // we need in this situation.
-func KeyToHash[K any](key K) (uint64, uint64) {
+func KeyToHash[K Key](key K) (uint64, uint64) {
 	keyAsAny := any(key)
```
mangalaman93, line 38: you probably don't need to do this anymore

pete-woods: Unfortunately, that's against the Go language spec (see generics proposal): https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md#why-not-permit-type-assertions-on-values-whose-type-is-a-type-parameter

TL;DR, you'll get an error `cannot use type switch on type parameter value` when trying to perform a type switch on a parameterized type.

mangalaman93 (approved):

pete-woods:
Thanks for your time! 🙇

PR #375, opened 2024-02-21
Fix build with GOOS=js GOARCH=wasm

paralin:
GOOS=js GOARCH=wasm gb ./...

Previously failed with errors importing x/sys/unix.

Fixed by adding the appropriate build tags.


paralin:
Still relevant

harshil-goel (approved):

PR #395, opened 2024-09-18
Use a min heap to reduce the time complexity while adding a new item

xiaoqingwanga:
Use a min heap to reduce the time complexity in the process of finding the minimum while adding a new item.
It seems fine, all tests have passed.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -17,7 +17,6 @@
 package ristretto
 
 import (
-	"math"
 	"sync"
 	"sync/atomic"
 
@@ -94,6 +93,11 @@ func (p *defaultPolicy[V]) CollectMetrics(metrics *Metrics) {
 type policyPair struct {
 	key  uint64
 	cost int64
+	hits int64
+}
+
+func (p policyPair) Less(other *policyPair) bool {
+	return p.hits <= other.hits
 }
 
 func (p *defaultPolicy[V]) processItems() {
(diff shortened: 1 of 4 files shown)


policy.go
```diff
@@ -129,45 +133,33 @@ func (p *defaultPolicy[V]) Add(key uint64, cost int64) ([]*Item[V], bool) {
 	// incHits is the hit count for the incoming item.
 	incHits := p.admit.Estimate(key)
 	// sample is the eviction candidate pool to be filled via random sampling.
-	// TODO: perhaps we should use a min heap here. Right now our time
-	// complexity is N for finding the min. Min heap should bring it down to
-	// O(lg N).
-	sample := make([]*policyPair, 0, lfuSample)
+	sample := NewMinHeap[policyPair]()
```
gameformush, line 136: Just passing by here 
any reasons to use custom min heap implementation instead of [std](https://pkg.go.dev/container/heap)?

Second point: If the goal of this change is to improve performance, it would be helpful to have benchmark results.

xiaoqingwanga:
I have conducted some benchmarks with the updated policy and the policy without any change:
updated policy:
image
policy without any change:
image
There  is a dramatic performance degradation. Building a min heap in policy adding procedure doesn't seem to be a good idea. A better idea needs to be came up with.

Issue #378, opened 2024-04-12
Cache configuration checks should consider negative values for `NumCounters`/`MaxCost`/`BufferItems`


ankon:
Found while reading through the code:
https://github.com/dgraph-io/ristretto/blob/c67e07994f911dbf42104f568c83a29830e88236/cache.go#L169-L174

The checks are fine, but given that these are `int64`s they could be <0 which the checks won't catch unless I'm missing something.

zhaohaihang:
I also think that the aforementioned variables should be set to be greater than 0.  Because if the aforementioned variables are less than 0, it will cause the program to fail.  Since we're already performing checks, why not do it even better?

mangalaman93:
Fixed by #410

PR #410, opened 2024-10-28
Perform validation to ensure that the three parameters, NumCounters, …

zhaohaihang:
…MaxCost, and BufferItems, are all greater than 0.(#378)


cache.go
```diff
@@ -166,12 +166,12 @@ type Item[V any] struct {
 // NewCache returns a new Cache instance and any configuration errors, if any.
 func NewCache[K Key, V any](config *Config[K, V]) (*Cache[K, V], error) {
 	switch {
-	case config.NumCounters == 0:
```
mangalaman93, line 169: Could you update this such that we do not change existing errors? This would mean keeping the existing code as is and introducing new cases and new errors.

zhaohaihang: I'm very sorry, I may have overlooked this case, and I will revise my PR.

mangalaman93 (approved):

mangalaman93:
Thank you for the PR @zhaohaihang

PR #411, opened 2024-10-30
set min version to go 1.21 in go.mod

mangalaman93:


harshil-goel (approved):

x1nchen:
@mangalaman93 hi, why did you set min version to go 1.21? Are there any issues with using go 1.20 version?

PR #425, opened 2024-12-17
refactor: using defer m.Unlock in store.go

hoo47:
- AS-IS: calls m.Unlock() before the end of function(method)
- TO-BE: using `defer` keyword, automatically calling unlock method right before return statements

### DIFF
--- a/store.go
+++ b/store.go
@@ -236,5 +236,4 @@ func (m *lockedMap[V]) Clear(onEvict func(item *Item[V])) {
 		}
 	}
 	m.data = make(map[uint64]storeItem[V])
-	m.Unlock()
 }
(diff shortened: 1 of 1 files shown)


store.go
```diff
@@ -223,12 +220,12 @@ func (m *lockedMap[V]) Update(newItem *Item[V]) (V, bool) {
 		expiration: newItem.Expiration,
 	}
 
-	m.Unlock()
 	return item.value, true
 }
 
 func (m *lockedMap[V]) Clear(onEvict func(item *Item[V])) {
 	m.Lock()
+	defer m.Unlock()
```
mangalaman93, line 228: Remove the m.Unlock() call at the bottom of this function

mangalaman93 (changes requested):

mangalaman93 (approved):

PR #427, opened 2024-12-25
Add ShouldUpdate in config

harshil-goel:
Add ShouldUpdate which allows users to check if value should be updated in the cache or not

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -148,8 +148,9 @@ type Config[K Key, V any] struct {
 	OnExit func(val V)
 
 	// ShouldUpdate is called when a value already exists in cache and is being updated. In this function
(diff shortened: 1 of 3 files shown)


cache.go
```diff
@@ -147,6 +147,11 @@ type Config[K Key, V any] struct {
 	// as well as on rejection of the value.
 	OnExit func(val V)
 
+	// ShouldUpdate is called when a value already exists in cache and is being updated. In this function
+	// you can check if the new value is valid or not. You can return a true or false is you want to continue to
```
mangalaman93, line 151: If*

Should we add an example as well?

mangalaman93 (dismissed):

mangalaman93 (approved):

shivaji-kharse (approved):

PR #429, opened 2024-12-26
fix memory leak on incor calc of cleanup bucket

flymedllva:
cleanup now does not try to clean an empty bucket selected by the cleanup algorithm

this fixes a critical problem of slice growth when the algorithm incorrectly calculates the buckets that need to be cleared, let's say this behavior occurs at the time change described in issue: https://github.com/dgraph-io/ristretto/issues/426

this problem appeared after improving cleanup in this pr: https://github.com/dgraph-io/ristretto/pull/358

most likely the algorithm for calculating the cleaning bins also needs to be revised, but for me now it is important to fix the increase in memory consumption when changing the system time

### DIFF
--- a/ttl.go
+++ b/ttl.go
@@ -121,9 +121,9 @@ func (m *expirationMap[_]) del(key uint64, expiration time.Time) {
 // cleanup removes all the items in the bucket that was just completed. It deletes
 // those items from the store, and calls the onEvict function on those items.
 // This function is meant to be called periodically.
-func (m *expirationMap[V]) cleanup(store store[V], policy *defaultPolicy[V], onEvict func(item *Item[V])) {
+func (m *expirationMap[V]) cleanup(store store[V], policy *defaultPolicy[V], onEvict func(item *Item[V])) int {
 	if m == nil {
-		return
+		return 0
 	}
 
 	m.Lock()
@@ -164,6 +164,10 @@ func (m *expirationMap[V]) cleanup(store store[V], policy *defaultPolicy[V], onE
 			}
 		}
 	}
+
+	cleanedBucketsCount := len(buckets)
+
+	return cleanedBucketsCount
 }
 
 // clear clears the expirationMap, the caller is responsible for properly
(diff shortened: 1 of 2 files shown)


mangalaman93 (approved):

mangalaman93:
@flymedllva Thank you for the PR. The ci passes. is it possible to write a test for the issue?

flymedllva:
I couldn't find a direct test for cleanup, so I made one and included in its implementation a sub-check that caused the slice to grow uncontrollably without allowing cleanup to terminate. With the fix, this test passes.

mangalaman93 (approved):

mangalaman93:
The test seems to fail:
```
--- FAIL: TestExpirationMapCleanup (5.00s)
    --- FAIL: TestExpirationMapCleanup/Miscalculation_of_buckets_does_not_cause_memory_leaks (1.00s)
        ttl_test.go:83: 
            	Error Trace:	/home/runner/work/ristretto/ristretto/ttl_test.go:83
            	Error:      	cleanup method hangs / there may be a memory leak!
            	Test:       	TestExpirationMapCleanup/Miscalculation_of_buckets_does_not_cause_memory_leaks
FAIL
```

flymedllva:
Apparently CI runners have fewer CPUs and because of this the tests take longer than on my computer. To fix this I had to introduce a cleanedBucketsCount variable in cleanup return, which I can use in the test to count the buckets cleaned by cleanup.

The test is no longer CPU-dependent and takes less time to run + it catches this case immediately during the first cleanup without reaching the test case with system time change (although it works too).

mangalaman93:
Thank you for updating the PR. The linter is not happy now:
```
Running [/home/runner/golangci-lint-1.62.2-linux-amd64/golangci-lint run --new-from-patch=/tmp/tmp-1914-PGn62ZkvM864/pull.patch --new=false --new-from-rev= --timeout=10m] in [/home/runner/work/ristretto/ristretto] ...
  Received 174177 of 174177 (100.0%), 0.2 MBs/sec
  ttl_test.go:72: the line is 124 characters long, which exceeds the maximum of 120 characters. (lll)
  		require.Equal(t, 0, cleanedBucketsCount, "cleanedBucketsCount should be 0 after cleanup with lastCleanedBucketNum change")
```

mangalaman93 (approved):

mangalaman93:
Thank you so much for the quick fix and addressing all the issues. Appreciate it.

PR #442, opened 2025-02-24
Ffz/use channel

freeformz:
**Description**

Two changes, somewhat related.

1. Switch from using a sync.Waitgroup, to closing a channel of `struct{}`. The later should be much more lightweight.
2. Aligned (using betteralign) some of the higher volume types, namely:

* Item: 32 bytes saved: struct with 80 pointer bytes could be 48
* storeItem: 16 bytes saved: struct with 56 pointer bytes could be 40
* shardedMap: 16 bytes saved: struct with 32 pointer bytes could be 16

**Checklist**

- [x] Code compiles correctly and linting passes locally (there is no linting instructions in the README, no golangci.yaml file, etc)
- [ ] For all _code_ changes, an entry added to the `CHANGELOG.md` file describing and linking to
      this PR
- [ ] ~~Tests added for new functionality, or regression tests for bug fixes added as applicable~~
- [ ] ~~For public APIs, new features, etc., PR on~~
      ~~[docs repo](https://github.com/dgraph-io/dgraph-docs) staged and linked here~~

**Instructions**

- The PR title should follow the [Conventional Commits](https://www.conventionalcommits.org/)
  syntax, leading with `fix:`, `feat:`, `chore:`, `ci:`, etc.
- The description should briefly explain what the PR is about. In the case of a bugfix, describe or
  link to the bug.
- In the checklist section, check the boxes in that are applicable, using `[x]` syntax.
  - If not applicable, remove the entire line. Only leave the box unchecked if you intend to come
    back and check the box later.
- Delete the `Instructions` line and everything below it, to indicate you have read and are
  following these instructions. 🙂

Thank you for your contribution to Ristretto!


store.go
```diff
@@ -14,10 +14,10 @@ type updateFn[V any] func(cur, prev V) bool
 
 // TODO: Do we need this to be a separate struct from Item?
 type storeItem[V any] struct {
+	expiration time.Time
+	value      V
```
mangalaman93, line 18: Here, value V should be at the end because it is variable length.

store.go
```diff
@@ -14,10 +14,10 @@ type updateFn[V any] func(cur, prev V) bool
 
 // TODO: Do we need this to be a separate struct from Item?
 type storeItem[V any] struct {
+	expiration time.Time
```
mangalaman93, line 17: How does moving expiration first help here?

store.go
```diff
@@ -55,8 +55,8 @@ func newStore[V any]() store[V] {
 const numShards uint64 = 256
 
 type shardedMap[V any] struct {
-	shards    []*lockedMap[V]
 	expiryMap *expirationMap[V]
+	shards    []*lockedMap[V]
```
mangalaman93, line 59: Please elaborate here as well and add a comment why this sequence is better

mangalaman93 (changes requested):

store.go
freeformz: I dropped the alignment changes from this PR.
I need to see what affect generics has on struct alignment before proposing them again.

mangalaman93 (approved):

PR #445, opened 2025-03-20
fix: Remove zerovalue[V].

aryehlev:
When i added generics in my PR i put the zerovalue to return valid values(since not all values will have nil).
a more elegant solution is just to have paramater returns in the function

**Checklist**

      this PR
- [x] For public APIs, new features, etc., PR on
      [docs repo](https://github.com/dgraph-io/dgraph-docs) staged and linked here

**Instructions**

- The PR title should follow the [Conventional Commits](https://www.conventionalcommits.org/)
  syntax, leading with `fix:`, `feat:`, `chore:`, `ci:`, etc.
- The description should briefly explain what the PR is about. In the case of a bugfix, describe or
  link to the bug.
- In the checklist section, check the boxes in that are applicable, using `[x]` syntax.
  - If not applicable, remove the entire line. Only leave the box unchecked if you intend to come
    back and check the box later.
- Delete the `Instructions` line and everything below it, to indicate you have read and are
  following these instructions. 🙂

Thank you for your contribution to Ristretto!

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -27,11 +27,6 @@ var (
 
 const itemSize = int64(unsafe.Sizeof(storeItem[any]{}))
 
-func zeroValue[T any]() T {
-	var zero T
-	return zero
-}
-
 // Key is the generic type to represent the keys type in key-value pair of the cache.
 type Key = z.Key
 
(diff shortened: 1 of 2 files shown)


mangalaman93:
Thank you for the PR @aryehlev. I do agree that not having zeroValue function is useful but at the same time, I do not like using named return. It reduces the readability of the code. Is there a way we could have both?

aryehklein-rise:
@mangalaman93 hey, the point was to get rid of zeroValue. 
i don't like it, named return solves that.
but i will close PR if u don't like it.

freeformz:
An alternative is to do this at the top of the function and then return that value....

```go
var value V
...
value = ....
....
return value, err
```

I tend to use this pattern extensively when writing generic Go code.

+1 on not using blank returns.
PS: You don't have to use blank returns with named returned values, you can still `return value, err`

mangalaman93:
> An alternative is to do this at the top of the function and then return that value....
> 
> ```go
> var value V
> ...
> value = ....
> ....
> return value, err
> ```
> 
> I tend to use this pattern extensively when writing generic Go code.
> 
> +1 on not using blank returns. PS: You don't have to use blank returns with named returned values, you can still `return value, err`

Doing this is same as what we have today, don't you think?

PR #448, opened 2025-04-11
Add public Cache.Cap() method

xl-s:
**Description**

Currently, there's no way to deduce how much of the max capacity is being used in a cache instance. This adds a new `Cap()` method that enables this. Some cases where this is useful:

* Monitoring the cache usage trend to facilitate better cache sizing
* Decide whether to apply backpressure to data fetching based on the available capacity

**Checklist**

      this PR

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -489,6 +489,14 @@ func (c *Cache[K, V]) UpdateMaxCost(maxCost int64) {
 	c.cachePolicy.UpdateMaxCost(maxCost)
 }
 
+// Cap returns the remaining cost capacity (MaxCost - Used) of an existing cache.
+func (c *Cache[K, V]) Cap() int64 {
+	if c == nil {
+		return 0
+	}
+	return c.cachePolicy.Cap()
+}
+
 // processItems is ran by goroutines processing the Set buffer.
 func (c *Cache[K, V]) processItems() {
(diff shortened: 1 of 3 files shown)


cache.go
```diff
@@ -489,6 +489,14 @@ func (c *Cache[K, V]) UpdateMaxCost(maxCost int64) {
 	c.cachePolicy.UpdateMaxCost(maxCost)
 }
 
+// Cap returns the remaining cost capacity (MaxCost - Used) of an existing cache.
+func (c *Cache[K, V]) Cap() int64 {
```
mangalaman93, line 493: Can we call it RemainingCost()?

cache_test.go
```diff
@@ -121,6 +121,39 @@ func TestUpdateMaxCost(t *testing.T) {
...
+		NumCounters: 10,
+		MaxCost:     100,
+		BufferItems: 64,
+	})
+	require.NoError(t, err)
+	require.Equal(t, int64(100), c.Cap())
+
+	// Set allowed into cache
+	require.True(t, c.Set(1, 1, 1))
+	time.Sleep(wait)
```
mangalaman93, line 135: Can we use c.Wait() here instead?

mangalaman93 (changes requested):

xl-s:
Updated it

PR #455, opened 2025-06-28
feat: Add Cache.MarshalBinary() and Cache.UnmarshalBinary([]byte)

cmmoran:
**Description**

Add Cache.MarshalBinary() and Cache.UnmarshalBinary([]byte) and shardedMap.snapshotEntries() to support Cache.MarshalBinary()

Currently, there is no way to export the cache or load the cache from exported data. This adds the same functionality with the intent of being able to export the cache to an external source such as a replication node or to facilitate cache warming.

Effort was made to ensure concurrency, TTLs, cost, and internal structure were preserved. Test cases for all new functions have been included. README.md was updated with the caveat and considerations that must be taken in order for MarshalBinary() to work properly; this is because MarshalBinary() (and by extension UnmarshalBinary([]byte)) use gob encoding/decoding under the hood. Comments have been added to the relevant new functions. The entire test suite has been executed to ensure sanity.

Also, the spirit (content?) of PR (#354) have been included because they represent a non-zero performance improvement. The one caveat to (#354) is this: it only works if `numShards` is a power of 2. With this in mind, a compile-time check has been added to ensure that `numShards` is a power of 2. Additionally, the compile-time check has comments instructing that the compile-time check must be removed AND `key&shardMask` must be changed back to `key%numShards` should `numShards` need to be changed to a value that is _not_ a power of 2.

Please explain the changes you made here.

**Checklist**

      this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ adhere to [Semantic Versioning](http://semver.org/spec/v2.0.0.html) starting v1.
 
 ## [Unreleased]
 
+**Added**
+
+- Add public Cache.MarshalBinary() and Cache.UnmarshalBinary([]byte) methods (#455)
+
 **Changed**
 
 - Remove dependency: github.com/pkg/errors (#443)
--- a/README.md
+++ b/README.md
@@ -94,6 +94,46 @@ func main() {
 }
 ```
 
+### ⚠️ Interface Types and `gob` Registration
+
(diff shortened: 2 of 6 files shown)


cmmoran:
Well, this is rather embarrassing. While this works on a single machine (within the same process) it will not work on different machines owing to ristretto's key hashing guarantee.

Closing.

cmmoran:
It appears that "simply" replacing xxhash with murmur3 (Sum64 vs Sum128) kills more than a couple birds with one stone.

1. It would remove ristretto from the [memhash wall of shame](https://github.com/golang/go/blob/b51f1cdb8752c2fdd6a2bf06224aab0b5c07caac/src/runtime/alg.go#L55)
2. It would solve the "TODO" item for z.KeyToHash regarding generating a single hash (128 bits) and using that to return both a key hash and a conflict hash.
3. Selfishly, it allows me to reopen this PR...

I'll hold off on reopening this PR unless there is some other interest in it.

PR #460, opened 2025-07-18
feat: implement iteration over cache elements

ilyatotl:
**Description**

Often, ***Ristretto*** is used not only as a cache before databases, but also as a complete in-memory storage. And there are cases when you need to do something with all the elements in the cache (for example, completely unload them and send them somewhere). However, in the current implementation, this is impossible. For this, the [Iter](https://github.com/hypermodeinc/ristretto/pull/460/files#diff-ed648181f98484bd12541509e0ae7b5ad1d1de7674ab1721814b47ddd5c95de4R87)  function was implemented, which goes through all the cache elements and calls a callback function for each one of them.

P.S. In addition, when running the tests, it was noticed that they were running for quite a long time, so they were parallelized, which significantly accelerated their work.

**Checklist**

      this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ adhere to [Semantic Versioning](http://semver.org/spec/v2.0.0.html) starting v1.
 
 - Remove dependency: github.com/pkg/errors (#443)
 - Add public Cache.RemainingCost() method
+- Implement public Cache.Iter() method
+- Make tests concurrent
 
 **Fixed**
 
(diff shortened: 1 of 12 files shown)


mangalaman93 (changes requested):
Thank you for the PR. Do we need to change the API here at all for having iteration function? If not, I'd recommend breaking this PR into two. It would be easy to get one in that does not change the API, and only adds a new API. We can later on review the other one if and when we plan for a major release.

PR #466, opened 2025-09-16
chore: typo in cost function

noahfraiture:
**Description**

Correct a small typo in Cost function parameter of new cache `whentthe` -> `when the`


noahfraiture:
It seems to be a problem with go version in the CI

Issue #483, opened 2026-03-24
[FEATURE] Split global policy mutex into separate locks for tinyLFU and sampledLFU


huynhanx03:
## Is your feature request related to a problem? Please describe

`defaultPolicy` uses a single `sync.Mutex` shared between `tinyLFU` (admission) and `sampledLFU` (eviction). All operations - `Add`, `Del`, `Has`, `Cap`, `Update`, `Cost`, `Clear`, and `processItems` - contend on this one lock.

This causes unnecessary contention:

- **`processItems` blocks on `Add`**: The `Add` method holds the lock for the entire eviction loop (sampling, estimating, deleting victims). During this time, `processItems` cannot push frequency counters to `tinyLFU`.
- **Reads block on writes**: `Has`, `Cap`, and `Cost` only read `sampledLFU.keyCosts` but acquire an exclusive lock, blocking all other operations.
- **Independent subsystems share a lock**: Frequency counter updates (`tinyLFU.Push`) and eviction metadata updates (`sampledLFU.add/del`) rarely need mutual exclusion.

## Describe the solution you'd like

Split into two locks:

```go
type defaultPolicy[V any] struct {
    admitMu  sync.Mutex   // protects tinyLFU
    evictMu  sync.RWMutex // protects sampledLFU
    admit    *tinyLFU
    evict    *sampledLFU
    // ...
}
```

**What changes:**

| Operation | Before | After |
|-----------|--------|-------|
| `processItems` (Push) | `Lock` (global) | `admitMu.Lock` only |
| `Has`, `Cap`, `Cost` | `Lock` (global) | `evictMu.RLock` |
| `Del`, `Update` | `Lock` (global) | `evictMu.Lock` |
| `Add` (eviction loop) | `Lock` (global, held entire loop) | `evictMu.Lock` + brief `admitMu.Lock` for `Estimate` |

**Lock ordering**: `admitMu` before `evictMu` (or never hold both) to prevent deadlocks.

## Describe alternatives you've considered

1. **Single `sync.RWMutex`**: Helps reads but `processItems` and `Add` still contend on write lock.
2. **Reduce critical section in `Add`**: Release/re-acquire between iterations. Adds complexity and potential inconsistency.
3. **Lock-free tinyLFU via atomics**: 4-bit counters in count-min sketch make atomics awkward.

## Additional context

- `processItems` already runs in its own goroutine, so it naturally benefits from a separate lock.
[...]

PR #489, opened 2026-06-27
perf: split admission and eviction locks

huynhanx03:
## Summary

Split the policy lock into independent admission and eviction locks. This prevents eviction work in `Add` from blocking TinyLFU frequency updates in `processItems` and allows eviction metadata reads to run concurrently.

## Changes

- Protect TinyLFU with `admitMu` and sampledLFU with `evictMu`.
- Release the eviction lock before frequency estimation and validate sampled victims before deletion.
- Add concurrency tests and focused policy benchmarks.

## Benchmark

CPU: Apple M5 (darwin / arm64)

| Name | Before | After | Change |
|---|---:|---:|---:|
| ProcessItems | 301.1 ns/op | 311.4 ns/op | +3.4% |
| ProcessItemsDuringEviction | 47,031 ns/op | 431.5 ns/op | -99.1% |
| Frequency throughput during eviction | 85,050 keys/s | 9,270,112 keys/s | 109x |
| Concurrent Has | 164.6 ns/op | 122.9 ns/op | -25.3% |
| Concurrent Cost | 145.2 ns/op | 103.4 ns/op | -28.8% |
| Concurrent Cap | 102.1 ns/op | 77.43 ns/op | -24.2% |
| Add with eviction | 908.1 ns/op | 1,079 ns/op | +18.8% |

## Checklist

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -24,7 +24,8 @@ func newPolicy[V any](numCounters, maxCost int64) *defaultPolicy[V] {
 }
 
 type defaultPolicy[V any] struct {
-	sync.Mutex
+	admitMu  sync.Mutex
+	evictMu  sync.RWMutex
 	admit    *tinyLFU
 	evict    *sampledLFU
 	itemsCh  chan []uint64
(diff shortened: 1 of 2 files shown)


huynhanx03:
@matthewmcneely This PR should resolve #483. Could you please take a look when you have a chance?

